PR #38, opened 2017-11-22
CDMD and clean code

ndem0:
- add the compress dmd
- remove the possibility of second matrix as input parameter for `fit` method
- add some internal method to avoid duplications

### DIFF
--- a/pydmd/dmdbase.py
+++ b/pydmd/dmdbase.py
@@ -22,14 +22,20 @@ class DMDBase(object):
 		is 0, that means no truncation.
 	:param bool exact: flag to compute either exact DMD or projected DMD.
 		Default is False.
-	:param original_time: dictionary that contains information about the time
-		window where the system is sampled: `t0` is the time of the first input
-		snapshot, `tend` is the time of the last input snapshot and `dt` is the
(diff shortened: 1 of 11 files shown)


pydmd/dmdbase.py
```diff
@@ -122,18 +121,27 @@ def reconstructed_data(self):
...
-			'Subclass must implement abstract method {}.fit'.
-			format(self.__class__.__name__)
+			'Subclass must implement abstract method {}.fit'.format(
+				self.__class__.__name__
+			)
 		)
 
-	def _fit_read_input(self, X, Y=None):
+	@staticmethod
+	def _col_major_2darray(X):
```
mtezzele: Fix the doc since the parameters are changed. Same for the returned values.

pydmd/dmdbase.py
```diff
@@ -225,6 +226,34 @@ def _compute_svd(X, svd_rank):
 
 		return U, s, V
 
+	@staticmethod
+	def _build_lowrank_op(U, s, V, Y):
```
mtezzele: Documentation. Same below.

PR #60, opened 2018-03-01
Add future to dependencies in readme

robbystk:
I installed via setup.py without problems but python complained on import about missing the `past` module which is part of the `future` package.


mtezzele:
@robbystk thank you very much for noting it and for the PR!

PR #68, opened 2018-05-15
DMD with control

ndem0:
- [x] implement `DMDc`
- [x] tests for `DMDc`
- [x] documentation for `DMDc`


mtezzele:
Just a test for the property btilde is missing.

PR #103, opened 2019-10-15
optdmd.py - missing return self in fit method

shervinsahba:
I believe the fit method is missing the return self call. I noticed this when running Tutorial 8, which threw errors but now works after making this fix. (PS Thanks for making this package!)


mtezzele:
@shervinsahba thank you very much for the contribution!

Issue #107, opened 2019-10-24
Ill conditioned data matrices - Checks and Regularization


adam2392:
Hi,

so I ran into an issue w/ my data matrices being fed in and there is are numerical issues when your X is ill-conditioned (e.g. > 10^4), then the SVD approach will surely fail. My runs give me basically useless A matrices.

Had two questions:

1. Any suggestions on how we might incorporate regularization into the code elegantly? I'm thinking something like L2 regularization should be sufficient, with a passed in parameter.
2. I plan on incorporating a few lines just to check the condition number of the X matrix. Any concerns on me doing this? Thinking something along the lines of 

"if cond > THRESHOLD: raise warning."

mtezzele:
#108 addresses point 2. 
For what concerns point 1 I would add something in the `_col_major_2darray(X)` method. Are you aware of some literature doing regularization before applying DMD? In case we can refer to that I suppose.

PR #108, opened 2019-10-24
Adding a condition number checker to raise warnings if it is so.

adam2392:
This would address issues where potentially a data matrix is ill-conditioned. Can happen when data is not normalized properly, but it would at least alert the user that such a situation occurred.

It passes all unit tests, and also includes a warning statement that suggests alternatives:
- preprocess data (normalization, etc.)
- possibly augmenting data matrix with stacking 
- apply regularization (this would be nice to implement?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,6 @@
 *.pyc
 pydmd\.egg-info/
 
+.idea/*
+
 tutorials/\.ipynb_checkpoints/
--- a/pydmd/dmd.py
+++ b/pydmd/dmd.py
@@ -1,7 +1,6 @@
 """
 Derived module from dmdbase.py for classic dmd.
 """
-import warnings
 
 # --> Import standard python packages
 import numpy as np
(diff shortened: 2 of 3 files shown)


adam2392:
#107 is partially addressed here.

adam2392:
I don't know if necessary to add to the rest of the fit functions, or just here. 

If so, perhaps I could migrate this into function "_col_major_2darray(X)"?

mtezzele:
Dear @adam2392 thank you very much for this PR. 
I think it is best to migrate this into the `_col_major_2darray(X)`method. Moreover, can you please make it work also for py2.7? We will end up the support for py2.7 in the near future but in the meantime we do our best ;)

adam2392:
@mtezzele migrated and fixed the warning string! (I think that works?) :p Moved over to py3.6+ awhile ago, so always iffy on the syntax.

mtezzele:
Thank you very much @adam2392 !

PR #121, opened 2020-02-16
Fix #120 through introducing original t0 in dynamics

### LINKED ISSUE #120: Incorrect `DMD.reconstructed_data` compared to `dmd.predict`
When `t0` (0 here) is a multiple of the period (3 here), `DMD.reconstructed` gives correct results
``` python
t = np.linspace(0, 5, 100)
X = np.sin(1 / 3 * 2 * np.pi * t).reshape((1, -1))  # frequency = 1 / 3
X_embedded = np.vstack([X[:, :-1], X[:, 1:]])

dmd = pydmd.DMD(svd_rank=2, exact=True, opt=True).fit(X_embedded)

dmd.original_time["dt"] = t[1] - t[0]
dmd.dmd_time["t0"] = t[0]
dmd.dmd_time["tend"] = t[-1]
dmd.dmd_time["dt"] = t[1] - t[0]

def dmd_predict(dmd, X0, n):
    Y = np.empty((X0.shape[0], n), dtype=np.complex)
    Y[:, 0] = X0
    for i in range(n - 1):
        Y[:, i + 1] = dmd.predict(Y[:, i])
    return Y

plt.plot(t, X[0], "k", alpha=0.5, linewidth=3, label="Original")
plt.plot(dmd.dmd_timesteps, dmd.reconstructed_data.real[0], "--", label="DMD reconstructed")
plt.plot(dmd.dmd_timesteps, dmd_predict(dmd, X_embedded[:, 0], len(dmd.dmd_timesteps)).real[0], "-.", label="DMD predicted")
plt.grid()
plt.legend()
plt.tight_layout()
```

Screenshot 2020-02-15 at 14 19 48

However when `t0 = 1`, `DMD.predict` is correct, not `DMD.reconstructed_data`

Screenshot 2020-02-15 at 14 20 51

When `t0 = 3` which is the signal period here, again `DMD.reconstructed_data` is correct

Screenshot 2020-02-15 at 14 21 50

We need to check in `DMD.reconstructed_data` how time translation affects the result.

tianyikillua:
It was implicitly assumed that the initial snapshot corresponds to `t = 0`. This PR brings meaning to the existing attribute `dmd.original_time["t0"]`which is now used when constructing the dynamics of the system.

The issue #120 is now fixed by this PR.

Unit tests are passing.

### DIFF
--- a/pydmd/dmdbase.py
+++ b/pydmd/dmdbase.py
@@ -124,8 +124,9 @@ class DMDBase(object):
         :rtype: numpy.ndarray
         """
         omega = old_div(np.log(self.eigs), self.original_time['dt'])
-        vander = np.exp(np.multiply(*np.meshgrid(omega, self.dmd_timesteps)))
-        return (vander * self._b).T
+        vander = np.exp(
+            np.outer(omega, self.dmd_timesteps - self.original_time['t0']))
+        return vander * self._b[:, None]
 
     @property
     def reconstructed_data(self):


ndem0:
@tianyikillua Thank you so much for the effort!

Issue #122, opened 2020-03-02
Fix for mrDMD division by zero error during reconstruction


rfmiotto:
Dear,

I am using PyDMD package to analyze the data from my research in fluid dynamics. Particularly, with the multi-resolution DMD (mrDMD), I was obtaining a zero-division error when doing the reconstruction of the flow field. This issue was reported by other users (see #81 , for example)

![zero-division-error](https://user-images.githubusercontent.com/35971460/75714623-0feb0180-5c9a-11ea-89fe-b0e2cebfd58a.png)

## Debugging

I noticed that the problem was coming from the exception of the `fit` method. That is, in the current implementation, if the number of resolution levels (used-defined parameter) is large, the `fit` method would return an error. In order to fix this problem, a try-except approach was used, and the eigenvalues, DMD modes, etc. were set as zero if the exception was raised. These null values were causing the problem in the reconstruction.

I solved this problem by proposing a small refactor in the `fit` method and I would like to make a pull request with the solution.

Moreover, I noticed that at the end of the `fit` method, the `dmd_time` and `original_time` dictionaries were being overwritten. It seemed to me that this is because the mrDMD cannot handle predictions yet (as mentioned in #63). So the `dmd_time` was forced to be equal to the `original_time`. 

Finally, a MemoryError exception was added for when reconstructing the solutions built from large matrices and with many resolution levels.

## Possible future contributions

In my research, I am interested in isolating a given physical phenomenon that occurs at a specific time interval. Since the dynamics have different spatiotemporal scales, the mrDMD came in handy. I had to implement more stuff in the MrDMD class in order to extract the modes related to the isolated temporal phenomenon. I believe that it might help other users and so, I will be pleased to include this contribution to the main source. Unit tests of these new methods will also come along.

Please, let me know what do you think of it.

Best regards,
Renato Miotto

ndem0:
Dear @rfmiotto, thank you for the effort! We'll check the Merge Request as soon as possible.

Regarding your dubts:
- the dictionaries initialization is performed for all the DMD classes, only to initialize the attributes (nothing related with the MrDMD prediction).
- if I remember correctly, the `try:...except` lines have been added in order to avoid errors that occur when the `step` variable is too large to fit in a single _window_. Better (and cleaner) solutions are very welcome!

rfmiotto:
Dear @ndem0,

Thank you for clarifying.

[...]

PR #123, opened 2020-03-02
Fix division by zero in MrDMD class

rfmiotto:
Dear,

As described in issue #122, here follows a proposal for fixing the division by zero in the multi-resolution DMD. The problem was being caused by the null eigenpairs in the `fit` method when the number of resolution levels defined by the user was large.

In the proposed pull request, a refactor was made only inside `fit` method. It is worth mentioning that this fix does not alter the solutions of the method, but avoids the zero-valued modes that were causing the problem in the reconstruction.

This pull request also includes a solution for the eventual `MemoryError` that raises when dealing with large matrices and large number of resolution levels. 

Best,
Renato Miotto

### DIFF
--- a/pydmd/mrdmd.py
+++ b/pydmd/mrdmd.py
@@ -60,7 +60,7 @@ class MrDMD(DMDBase):
         :rtype: int
         :return: the index of the list that contains the binary tree.
         """
-        if level > self.max_level:
+        if level >= self.max_level:
             raise ValueError("Invalid level: greater than `max_level`")
 
         if node >= 2**level:
@@ -186,7 +186,7 @@ class MrDMD(DMDBase):
         """
         if level >= self.max_level:
             raise ValueError(
-                'The level input parameter ({}) has to be less then the'
+                'The level input parameter ({}) has to be less than the'
                 'max_level ({}). Remember that the starting index is 0'.format(
                     level, self.max_level))
         if node:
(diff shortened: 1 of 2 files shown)


mtezzele:
Hi @rfmiotto! Thank you for your contribution. I kindly ask you to provide a couple of unit tests that check the while and the try except you added. Moreover at the end please squash the commits.

rfmiotto:
Dear @mtezzele, I squashed the commits and added a test for the maximum level, as requested.

I don't know if there is an efficient way to test for the `MemoryError` exception without creating a huge snapshot matrix. This approach can be very time consuming for the CI checks, I guess, so I dropped the idea.

**Note:** As you can see, some of the <ins>tests were changed</ins>. That was necessary in order to maintain the variables consistent. The code mimics the recursion process by adding and popping elements (the bins) into a queue during the while loop and <ins>not</ins> the level as stated before by the `current_level` variable. A new definition of the levels and `max_level` was adopted in this commit, and consequently, it was necessary to make some changes in the code, including the unit tests. The figure below tries to clarify how the levels are being defined now:

IMG_20200304_244930385
In this figure, the purple numbers inside the tree structure indicate the index of each bin. These are the values returned by the `_index_list` method. 

The resolution level starts at zero. Hence, when one sets `max_level = 3`, he is defining the levels to be [0, 1, 2, **3**], which contains 4 items.

Please, let me know if you need anything else, or if you have any questions about this solution.
Best regards

PR #124, opened 2020-03-04
Fix bug max lvl

rfmiotto:
Dear,

Sorry, I just found a bug in the evaluation of the maximum level threshold. It was fixed and I also included another assertion in the unit test to make sure it is OK.
I did a "fixup" of this commit with the previous one, and already fixed the conflicts that the CI complained about. 

Thank you,
Best,

### DIFF
--- a/tests/test_mrdmd.py
+++ b/tests/test_mrdmd.py
@@ -4,7 +4,7 @@ from unittest import TestCase
 from pydmd.mrdmd import MrDMD
 import matplotlib.pyplot as plt
 import numpy as np
-import unittest
+
 
 def create_data():
     x = np.linspace(-10, 10, 80)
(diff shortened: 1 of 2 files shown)


tests/test_mrdmd.py
```diff
@@ -42,8 +42,9 @@ def test_max_level_threshold(self):
         level = 10
         dmd = MrDMD(svd_rank=1, max_level=level, max_cycles=2)
         dmd.fit(X=sample_data)
-        lvl_threshold = int(np.log(sample_data.shape[1]/4.)/np.log(2.)) - 1
+        lvl_threshold = int(np.log(sample_data.shape[1]/4.)/np.log(2.)) + 1
         assert lvl_threshold == dmd.max_level
+        assert dmd._steps[-1] == 1
```
mtezzele: Please split the asserts in different tests. One assert per test.

rfmiotto: Hi @mtezzele, 

Done!

Best

mtezzele:
@rfmiotto you should rebase this branch. I would do it for you by web but I can't due to conflicts.

rfmiotto:
Hi @mtezzele,

Hmmm, I guess I did something that I think I shouldn't.
Previously, I have combined this commit into the previous one with "fixup" (after all, the commit in this PR should be part of the previous one). As a consequence, the commit in this PR ended up being the same as the one that already exists in the remote, causing the conflict. I have solved these conflicts from inside GitHub (that is why all checks have passed), but I believe that if you pull the code into your local machine, git will also complain about conflicts in you local repo. Is that what you said is happening?

Well, now I did something different to avoid these conflicts: I will leave this PR as a separate commit. I don't know how to squash/fixup the commit into the previous without causing these conflicts, so I will keep them separated. Please, let me know if you are aware of some workaround.

Thank you,
Best regards

mtezzele:
@rfmiotto now the git graph is perfect. I think the problem was that you merged master master into this one instead of rebasing wrt master.

PR #126, opened 2020-03-31
Add methods to extract the modes from a given time window

rfmiotto:
Dear,

This pull request includes a new feature into the mrDMD module: the possibility to extract modes within a given time window.

I am using this to analyse the modes of a transient dynamical systems (in which the dynamics is constantly changing from time to time), or of systems with intermittent events (in which a particular behavior only happens at given time).

The idea consists of finding the bins that are being "touched" by a user-defined time interval and extract the modes, amplitudes, frequencies and growth rate of these bins. The "time interval" is not an actual physical time, but the indexes of the snapshot corresponding to that physical time (because of the way `self.original_time` and `self.dmd_time` were implemented).

For example, say that we have the following collection of snapshots and that we are only interested in the event happening in the marked region:

Screenshot from 2020-03-31 08-02-37

We apply the mrDMD in such a way that, given the snapshot indexes corresponding to the ends of the window of interest (1, 4), we extract the modes whose bins are touched by it. The image below illustrates the idea. So all bins that are being touched by the blue area is going to be taken into account.

image

The motivation for this is totally dissociated from the reconstruction or mode reduction of the data, but aims to analyse the modes related to a specific event instead. If the dynamics is transient in nature, applying the standard DMD with a narrow time window to isolate the event might give modes that are difficult to be interpreted or even misleading. The mrDMD variant, however, can do a much better job in this sense, and in order to accomplish this, the methods present in this pull request had to be added.

Please, let me know what do you think. If necessary, I can provide more details.

Thank you,
Best regards
Renato Miotto

### DIFF
--- a/tests/test_mrdmd.py
+++ b/tests/test_mrdmd.py
@@ -51,6 +51,101 @@ class TestMrDmd(TestCase):
         dmd.fit(X=sample_data)
         assert dmd._steps[-1] == 1
 
+    def test_index_list(self):
+        level = 5
+        dmd = MrDMD(svd_rank=1, max_level=level, max_cycles=2)
+        assert dmd._index_list(3, 0) == 7
+
+    def test_index_list2(self):
+        level = 5
+        dmd = MrDMD(svd_rank=1, max_level=level, max_cycles=2)
+        with self.assertRaises(ValueError):
+            dmd._index_list(3, 10)
+
+    def test_index_list_reversed(self):
+        level = 3
+        dmd = MrDMD(svd_rank=1, max_level=level, max_cycles=2)
+        assert dmd._index_list_reversed(6) == (2, 3)
+
+    def test_index_list_reversed2(self):
+        level = 3
+        dmd = MrDMD(svd_rank=1, max_level=level, max_cycles=2)
+        with self.assertRaises(ValueError):
+            dmd._index_list_reversed(7)
+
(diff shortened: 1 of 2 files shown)


mtezzele:
Dear @rfmiotto, thank you again for your contributions. I see no problems in merging this PR.
If you have time maybe in the future you can consider to show these new functionalities in the mrDMD tutorial... ;)

tests/test_mrdmd.py
```diff
@@ -4,7 +4,7 @@
 from pydmd.mrdmd import MrDMD
 import matplotlib.pyplot as plt
 import numpy as np
-
+import unittest
```
mtezzele, line 7: is this necessary? I think we can also remove the `if __name__`... at the end.

rfmiotto: Oh, this is not necessary and you can remove it if you want. I put it there when I was running the tests on my computer, but forgot to remove it.

rfmiotto:
Hi @mtezzele ,

Sure, I can show how this functionality works in the mrDMD tutorial.
I will try to that ASAP ;)

Thank you,
Best

tests/test_mrdmd.py
rfmiotto: Just a heads up: I have just removed that from the PR  ;)

PR #134, opened 2020-10-14
Add tutorial-5-fbdmd.py

fpichi:


tutorials/tutorial-5-fbdmd.py
```diff
@@ -0,0 +1,69 @@
+import matplotlib.pyplot as plt
+import numpy as np
+import time
+
+from pydmd import FbDMD
+from pydmd import DMD
+
+from ipywidgets import interact_manual, interact, interactive, FloatSlider
```
mtezzele, line 8: I think this does not work on a script. Please check

fpichi: Actually I already tried successfully to run it before the PR. Moreover I think that the import of interact_manual and interact are indeed unnecessary also for the .ipynb, but I left them for consistency.

Issue #143, opened 2020-11-20
Adding a helper function to compute snapshots used in HODMD?


adam2392:
# Main issue
I was wondering if it would be okay with developers to add in a public function to HoDMD to recompute the snapshots?

E.g.

```Python
def get_snapshots():
        snaps = np.concatenate(
            [
                self._snapshots[:, i:self._snapshots.shape[1] - self.d + i + 1]
                for i in range(self.d)
            ],
            axis=0)
      return snaps
```

The reason is, I am trying to rescale the DMD modes computed and some of the algorithms for rescaling want to do underlying computations on the original snapshots ran through HoDMD. 

# Issue about allowing different scaling strategies
With that being said a small side issue I was having is how to rescale the modes. For example, in this paper on page 5, authors have scaled the modes by the singular values of the corresponding SVD, so the eigenvectors are `W = \Sigma^{1/2} \hat{W}` instead of `\hat{W}`, which is then used to compute the DMD modes. 

Would it be desirable to either i) add alternative scaling strategies on the DMD modes or ii) separate out the steps inside `fit()` a bit more such that, if someone wanted to alter say in between steps, they can start off there and call some private/public functions?

ndem0:
Dear @adam2392,
yes, we can easily add a function in the `HODMD` class for arranging the snapshots matrix.

But, actually I didn't get why this should help you in rescaling the DMD modes. Do you need a public function that returns the 'higher-order' snapshots matrix or a way to post-process the DMD modes?

adam2392:
Hi @ndem0 thanks for the response. Since posting this, I found a way to rescale the DMD modes by its corresponding singular values from the input data matrix according to the paper I linked.

```Python
def rescale_modes(snapshots, atilde, dmd):
    """Helper function to rescale and recompute DMD modes.

    Re-scales the DMD modes by the singular values from
    the data matrix SVD.

    Parameters
    ----------
    snapshots : np.ndarray
        The original snapshots matrix.
    atilde : np.ndarray
        The fitted rank-truncated linear operator from DMD.
    dmd : instance of pydmd.DMDBase
        ``dmd`` should already have been fitted.

    Returns
    -------
    modes : np.ndarray
        The re-scaled DMD modes.
    """
    Y = snapshots[:, 1:]

    # compute the SVD
    U, s, V = dmd._compute_svd(snapshots[:, :-1], dmd.svd_rank)

    # compute Ahat from Atilde
[...]

PR #151, opened 2021-03-01
rescale_mode parameter for dmdbase.py

fAndreuzzi:
I'm trying to implement the fix to #143 proposed by @ndem0. Before proceeding I would like to know if the position of the new parameter is OK, and if I'm using an acceptable documentation style.

### DIFF
--- a/pydmd/dmdbase.py
+++ b/pydmd/dmdbase.py
@@ -29,6 +29,9 @@ class DMDBase:
     :param bool exact: flag to compute either exact DMD or projected DMD.
         Default is False.
     :param bool opt: flag to compute optimized DMD. Default is False.
+    :param numpy.array rescale_mode: None means no rescaling, empty array means
+        automatic rescaling using SV, otherwise the user chooses the preferred
+        scaling.
     :cvar dict original_time: dictionary that contains information about the
         time window where the system is sampled:
 
@@ -44,7 +47,10 @@ class DMDBase:
             - `dt` is the delta time between the approximated solutions.
 
     """
-    def __init__(self, svd_rank=0, tlsq_rank=0, exact=False, opt=False):
+
+    def __init__(self, svd_rank=0, tlsq_rank=0, exact=False, opt=False,
+        rescale_mode=None):
+        self.rescale_mode = rescale_mode
         self.svd_rank = svd_rank
         self.tlsq_rank = tlsq_rank
         self.exact = exact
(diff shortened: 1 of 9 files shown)


pydmd/dmdbase.py
```diff
@@ -29,6 +29,9 @@ class DMDBase(object):
     :param bool exact: flag to compute either exact DMD or projected DMD.
         Default is False.
     :param bool opt: flag to compute optimized DMD. Default is False.
+    :param numpy.array rescale_mode: None means no rescaling, empty array means
+        automatic rescaling using SV, otherwise the user chooses the preferred
+        scaling.
```
ndem0, line 34: I would prefer to make clearer the automatic rescaling: instead of passing an empty array (which can be very problematic to debug in certain circumstances), I suggest to use a particular string (like `'auto'` or whatever). This also make easier to extend in future in the case other scaling become popular in the literature!

fAndreuzzi:
After a _superficial_ analysis of the code my opinion is that putting modifications in the static method `_eig_from_lowrank_op` is the best choice in order to affect all the subclasses of `DMDBase`. What do you think @ndem0 ?

fAndreuzzi:
Since the method
`@staticmethod
    def _eig_from_lowrank_op(Atilde, Y, U, s, V, exact, rescale_mode):`
is static, it is not affected if we set the class parameter _rescale_mode_ in _DMDBase_ unless we pass it as a parameter. This makes the class parameter _rescale_mode_ useless, and we should remove it. Is this the intended behavior for this new feature, @ndem0 ?

fAndreuzzi:
@ndem0 , following the idea proposed in #143 I added a missing line to scale eigenvectors before the computation of modes. I'm a little confused about the following line (from the function proposed **in the issue**):
```
modes = modes[:dmd._snapshots.shape[0], :]
```
It's apparent to me that this is done in order to select only _some_ modes, depending on choices made before the _first_ calculation of the modes. However, how should I behave if I'm calculating the modes for the first time? Which parameter is controlling the number of modes?

ndem0:
> @ndem0 , following the idea proposed in #143 I added a missing line to scale eigenvectors before the computation of modes. I'm a little confused about the following line (from the function proposed **in the issue**):
> 
> ```
> modes = modes[:dmd._snapshots.shape[0], :]
> ```
> 
> It's apparent to me that this is done in order to select only _some_ modes, depending on choices made before the _first_ calculation of the modes. However, how should I behave if I'm calculating the modes for the first time? Which parameter is controlling the number of modes?

It looks to me just a control to take a number of modes that is equal to the snapshots passed. Looking at the code it seems not particular useful (I can't imagine any case where the number of eigenvectors is higher than the number of snapshots) and I suggest to ignore it!

PR #155, opened 2021-03-17
Introduce class DMDOperator

fAndreuzzi:
This pull request introduces the new class `DMDOperator` in order to improve code readability. Some subclasses of `DMDBase` are already adapted to work with the new class (only `MrDMD` left). 

Several slight changes to tests have been made, but the public interface is mostly unchanged.


fAndreuzzi:
Tests are executing properly now, the conversion of `MrDMD` is done.

## TODO:
- There is a bit of confusion regarding to how to access and use `Atilde`. For instance, calling `self._Atilde.compute_operator(...)` in `fit()` is fine, but calling `self.atilde.compute_operator(...)` will raise an `Error`, since `atilde` is a "label" for `as_numpy_array`, which doesn't work when the operator isn't ready. I would like to get some suggestions on what names I should use, and what to expose to external users of the class `DMDOperator`;
- If we want to preserve the public interface, we should undo the changes I did in the `tests` folder;
- A huge review of the documentation, which I neglected while I was making changes on the code.

pydmd/dmdc.py
```diff
@@ -5,10 +5,47 @@
...
+            np.reciprocal(sp))).dot(Up1.T.conj()).dot(Ur)
+        self._compute_eigenquantities()
+        self._compute_modes_and_Lambda(Y, sp, Vp, Up1, Ur)
+
+        Btilde = Ur.T.conj().dot(Y).dot(Vp).dot(np.diag(
+            np.reciprocal(sp))).dot(Up2.T.conj())
+
+        return Ur, Ur.dot(Btilde)
+
+    def _compute_modes_and_Lambda(self, Y, sp, Vp, Up1, Ur):
```
mtezzele, line 45: naming convention: no capital please

pydmd/dmd.py
```diff
@@ -89,13 +83,14 @@ def predict(self, X):
...
         # --> Predict using the DMD modes as the basis.
         elif self.exact is True:
-            adjoint_modes = pinv(self._modes)
+            adjoint_modes = pinv(self.modes)
             Y = np.linalg.multi_dot(
-                [self._modes, np.diag(self._eigs), adjoint_modes, X]
+                [self.modes, np.diag(self.eigs), adjoint_modes, X]
             )
 
         return Y
```
ndem0, line 96: Now it should be just `return self._Atilde(X)`

pydmd/dmdoperator.py
```diff
@@ -0,0 +1,182 @@
...
+            scaling_factors_array = self._rescale_mode
+
+            factors_inv_sqrt = np.diag(np.power(scaling_factors_array, -0.5))
+            factors_sqrt = np.diag(np.power(scaling_factors_array, 0.5))
+            Ahat = np.linalg.multi_dot([factors_inv_sqrt, self.as_numpy_array,
+                factors_sqrt])
+
+        self._eigenvalues, self._eigenvectors = np.linalg.eig(Ahat)
+
+    def _compute_modes_and_Lambda(self, Y, U, Sigma, V):
```
mtezzele, line 158: I would refactor with something like: _compute_modes_and_eigenvalues or _compute_modes_and_eigs

pydmd/dmdbase.py
```diff
@@ -50,23 +51,21 @@ class DMDBase(object):
 
     """
 
-    def __init__(self, svd_rank=0, tlsq_rank=0, exact=False, opt=False,
-        rescale_mode=None):
-        self.rescale_mode = rescale_mode
-        self.svd_rank = svd_rank
+    def __init__(self, tlsq_rank=0, opt=False, **operator_kwargs):
```
ndem0, line 54: It's very dangerous change the order of arguments. Please revert and pass manually the arguments.

pydmd/dmdc.py
```diff
@@ -5,10 +5,47 @@
...
 """
-from .dmdbase import DMDBase
 from past.utils import old_div
 import numpy as np
 
+from .dmdbase import DMDBase
+from .dmdoperator import DMDOperator
+
+class DMDControlOperator(DMDOperator):
+    def __init__(self, svd_rank_omega=-1, **kwargs):
```
ndem0, line 15: Esplicit arguments.

pydmd/dmdc.py
```diff
@@ -5,10 +5,47 @@
...
+        super(DMDControlOperator, self).__init__(**kwargs)
+        self._svd_rank_omega = svd_rank_omega
```
ndem0, line 17: In the old version, it was `svd_rank_omega`, isn't? Please do not change the 'pubblic' attributes

pydmd/dmdc.py
```diff
@@ -5,10 +5,47 @@
...
+
+class DMDBKnownOperator(DMDControlOperator):
+    def compute_operator(self, X, Y, B, controlin):
+        Y = Y - B.dot(controlin)
+        return super(DMDBKnownOperator, self).compute_operator(X,Y)
```
ndem0, line 22: space in `compute_operator` arguments

pydmd/dmdc.py
```diff
@@ -39,26 +76,21 @@ class DMDc(DMDBase):
...
-        self.svd_rank = svd_rank
-        self.tlsq_rank = tlsq_rank
-        self.opt = opt
-        self.svd_rank_omega = svd_rank_omega
-        self.original_time = None
-        self.rescale_mode = rescale_mode
-
-        self._eigs = None
-        self._Atilde = None
+    def __init__(self, tlsq_rank=0, opt=False, **kwargs):
```
ndem0, line 79: explicit arguments

pydmd/dmdoperator.py
```diff
@@ -0,0 +1,182 @@
...
+        self._compute_modes_and_Lambda(Y, U, s, V)
+
+        return U, s, V
+
+    @property
+    def shape(self):
+        return self.as_numpy_array.shape
+
+    def __call__(self, matrix):
+        return self.Atilde.dot(matrix)
```
ndem0, line 37: What is supposed to be `matrix`? High or low dimensional snapshot?

pydmd/dmdoperator.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @property
+    def Lambda(self):
+        if not hasattr(self, '_Lambda'):
+            raise ValueError('You need to call fit before')
+        return self._Lambda
+
+    @property
+    def as_numpy_array(self):
+        return self._Atilde
```
ndem0, line 65: I'm confused: `_Atilde` is the object `DMDOperator` or the corresponding `numpy.ndarray`?

pydmd/dmdbase.py
```diff
@@ -50,23 +51,21 @@ class DMDBase(object):
 
     """
 
-    def __init__(self, svd_rank=0, tlsq_rank=0, exact=False, opt=False,
-        rescale_mode=None):
-        self.rescale_mode = rescale_mode
-        self.svd_rank = svd_rank
```
ndem0, line 56: Do not remove 'pubblic' attributes, for retrocompatibility. I think we can add easily solve by adding the `svd_rank` property that simply return the rank of the operator

tests/test_cdmd.py
```diff
@@ -46,7 +46,7 @@ def test_truncation_shape(self):
     def test_Atilde_shape(self):
         dmd = CDMD(svd_rank=3)
         dmd.fit(X=sample_data)
-        assert dmd.atilde.shape == (dmd.svd_rank, dmd.svd_rank)
+        assert dmd.atilde.shape == (3, 3)
```
ndem0, line 49: Do not change tests

tests/test_dmd.py
```diff
@@ -51,7 +51,7 @@ def test_rank(self):
     def test_Atilde_shape(self):
         dmd = DMD(svd_rank=3)
         dmd.fit(X=sample_data)
-        assert dmd.atilde.shape == (dmd.svd_rank, dmd.svd_rank)
+        assert dmd.atilde.shape == (3, 3)
```
ndem0, line 54: Do not change tests

tests/test_dmdbase.py
```diff
@@ -13,11 +13,11 @@
 class TestDmdBase(TestCase):
     def test_svd_rank_default(self):
         dmd = DMDBase()
-        assert dmd.svd_rank == 0
+        assert dmd._Atilde._svd_rank == 0
```
ndem0, line 16: Do not change tests

tests/test_hodmd.py
```diff
@@ -51,7 +51,7 @@ def test_rank(self):
     def test_Atilde_shape(self):
         dmd = HODMD(svd_rank=3)
         dmd.fit(X=sample_data)
-        assert dmd.atilde.shape == (dmd.svd_rank, dmd.svd_rank)
+        assert dmd.atilde.shape == (3, 3)
```
ndem0, line 54: Do not change tests

pydmd/dmdoperator.py
```diff
@@ -0,0 +1,182 @@
...
+
+        if self._rescale_mode == 'auto':
+            self._rescale_mode = s
+
+        self._Atilde = atilde
+        self._compute_eigenquantities()
+        self._compute_modes_and_Lambda(Y, U, s, V)
+
+        return U, s, V
+
```
ndem0, line 31: I would add a plot for the operator (mainly for debug) using https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.matshow.html

pydmd/dmdoperator.py
fAndreuzzi: I didn't design a specific usage for this method. Where should we use it?

pydmd/dmdoperator.py
fAndreuzzi: In `DMDOperator` `self` is `DMDOperator` and `self._Atilde` is a `numpy.ndarray`, in `DMDBase` `self._Atilde` is a `DMDOperator` and `self.atilde` is `self._Atilde.as_numpy_array` (so ´numpy.ndarray`). I know it's not good, this is way I was asking a suggestion on this topic in my comment above.

pydmd/dmdoperator.py
fAndreuzzi: I wrote `_compute_modes_and_Lambda` because `eigs` or `eigenvalues` could lead to misunderstandings (lowrank or high dimensional eigenvalues?).

Maybe something like `_compute_modes_and_hdeigs`? Or just `_compute_modes`, since high dimensional eigenvalues are the same of the low rank case.

pydmd/dmdoperator.py
ndem0: Just the operator definition: we pass a snapshots and it return the approximated one corresponding to next timestep (remember by definition we should have that `y = Ax`)

fAndreuzzi: So something like the `predict` method of `DMDBase`?

ndem0: Yep! We have two options:
- `predict` of `DMDbase` operates on high dimensional snapshots, while the `__call__` of `DMDOperator` manage the low-dimensional snapshots
- both method operate on high-dimensional snaps and so in the `predict` we just need to call the `__call__` method
The first should be the most versatile.

pydmd/dmd.py
fAndreuzzi: Please check aa5ec68

pydmd/dmdbase.py
fAndreuzzi: Do you recommend passing arguments manually only from `DMDBase` subclasses to `DMDOperator` subclasses, or everywhere?

ndem0: I strongly suggest to change everywhere.
It's more code, right, but it is also safer

pydmd/dmdoperator.py
mtezzele: Sure, last option I think is the best. Thanks

pydmd/dmdbase.py
fAndreuzzi: Fixed in f64e202

I removed the method `_initialize_dmdoperator` from `DMDBase` (and subclasses) since it wasn't needed anymore. I added it in order to decouple the instatiation of `DMDOperator` subclasses and setting class variables of `DMDBase` subclasses, therefore one could use a different `DMDOperator` implementation without having to touch the constructor. But this doesn't work unless we allow passing dicts, so I removed the whole structure.

pydmd/dmdc.py
fAndreuzzi: Fixed in c6de139

pydmd/dmdoperator.py
fAndreuzzi: Fixed in c6de139

pydmd/dmdc.py
fAndreuzzi: Fixed in f64e202

pydmd/dmdc.py
fAndreuzzi: Fixed in f64e202__

pydmd/dmdc.py
fAndreuzzi: Fixed in a479e0e

tests/test_hodmd.py
fAndreuzzi: Fixed in 82059f1

tests/test_dmdbase.py
fAndreuzzi: Fixed in 82059f1

tests/test_dmd.py
fAndreuzzi: Fixed in 82059f1

tests/test_cdmd.py
fAndreuzzi: Fixed in 82059f1

pydmd/dmdbase.py
fAndreuzzi: Fixed in 82059f1

pydmd/dmdc.py
fAndreuzzi: Fixed in 82059f1

pydmd/dmdoperator.py
fAndreuzzi: Please check cffd605

pydmd/dmdoperator.py
fAndreuzzi: Please check 5af1162

fAndreuzzi:
> There is a bit of confusion regarding to how to access and use Atilde. For instance, calling self._Atilde.compute_operator(...) in fit() is fine, but calling self.atilde.compute_operator(...) will raise an Error, since atilde is a "label" for as_numpy_array, which doesn't work when the operator isn't ready.
My proposal: 
- `DMDBase`:
   + `self._dmdoperator` is DMDOperator;
   + `self.atilde` is `numpy.ndarray` (the former `_Atilde`);
- `DMDOperator`:
   + `self.as_numpy_array` is `numpy.ndarray`.

What do you think @ndem0, @mtezzele?

ndem0:
> My proposal:
> 
> * `DMDBase`:
>   
>   * `self._dmdoperator` is `DMDOperator`;
>   * `self.atilde` is `numpy.ndarray` (the former `_Atilde`);
> * `DMDOperator`:
>   
>   * `self.as_numpy_array` is `numpy.ndarray`.
> 
> What do you think @ndem0, @mtezzele?

Yes, I was thinking to something similar!
I would propose only a modification:
`DMDBase` (and variants) with `self.operator` for the `DMDOperator` and `atilde` as before (the`ndarray`). Il looks more intuitive to call (e.g. `dmd = DMD().fit(X); dmd.operator.plot()`)

pydmd/dmdoperator.py
fAndreuzzi: Fixed in b2c6914

PR #159, opened 2021-03-23
DMD parameter 'amplitudes_snapshot_index'

fAndreuzzi:
In this commit I introduced the `DMDBase` parameter `amplitudes_snapshot_index`, which allows setting a time snapshot (which may not be the first one) to be used for the computation of DMD modes amplitudes. This results in a slight improvement of `reconstructed_data` in time instants near the chosen snapshot.

image

The two plots trepresent the error between the original function and the representation in `dmd.reconstructed_data` (time grows vertically). In the plot on the left I used the parameter `amplitudes_snapshot_index=0` (first snapshot), while in the plot on the right `amplitudes_snapshot_index=-1` (which means the last time snapshot).<br>As you can see in both cases the error vanishes as we approach the selected snapshot.

However the approach used has some limitations. When `Atilde` has small eigenvalues we observe a "blow-up" of the dynamics of the system (due to the negative power in the formula), which leads to wrong results. I observed that in some cases this is fixed by setting `svd_rank` to a value between 0 and 1, in order to select only "massive" eigenvalues.

### DIFF
--- a/pydmd/cdmd.py
+++ b/pydmd/cdmd.py
@@ -117,10 +117,21 @@ class CDMD(DMDBase):
     :param bool forward_backward: If True, the low-rank operator is computed
         like in fbDMD (reference: https://arxiv.org/abs/1507.02264). Default is
         False.
+    :param int amplitudes_snapshot_index: The (temporal) index of the snapshot
+        used to compute DMD modes amplitudes. The reconstruction will generally
+        be better in time instants near the chosen snapshot; however increasing
+        this value may lead to wrong results when the system presents small
+        eigenvalues. For this reason a manual selection of the number of
+        eigenvalues in the system may be needed (check svd_rank). Also setting
+        svd_rank to a value between 0 and 1 can lead to better results. Default
+        value is 0.
     """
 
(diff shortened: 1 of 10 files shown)


fAndreuzzi:
In [this](https://github.com/fAndreuzzi/PyDMD/blob/last-snapshot2/tutorials/blow-up.ipynb) notebook you can observe the problem. As you can see selecting the first 2 eigenvalues and modes fixes the reconstruction.

image

ndem0:
Great! I've only one concern regarding the new argument: now if the `opt` flag is set to True, the `amplitudes_snapshot_index` is ignored, and for sure we have to write it to the documentation.
Is not better to change the behavior of `opt` flag instead? Eg `opt=int` is the index of the snapshot for computing the amplitudes, `opt=True` perform the global optimization as before. I would prefer something like `opt='global'` for the latter case, but this means losing a bit of retrocompatibility.

fAndreuzzi:
> Is not better to change the behavior of opt flag instead?

Sure!

fAndreuzzi:
@ndem0 I made the changes you suggested, after your review I'm going to squash my commits and push again.

PR #171, opened 2021-04-27
Improve plot_eigs

fAndreuzzi:
In this pull request I introduce some changes to improve `DMDBase.plot_eigs`:
- `narrow_view` is now compatible with `show_axes=True` (you can see the result in the attached image);
- the ratio of the resulting image induced by `narrow_view` must be lower than 8 (to fix the fact that when, for instance, all the eigenvalues have the same complex part, the image degenerates to an horizontal line);
- `plt.legend(..., loc='best')` instead of `loc='1', which places the legend in the optimal place to reduce the overlap with the plot (instead of placing it always on the top right), the attached image demonstrates the behavior.

image

### DIFF
--- a/pydmd/dmdbase.py
+++ b/pydmd/dmdbase.py
@@ -514,6 +514,47 @@ class DMDBase(object):
             return partial(DMDBase.ModesSelectors._integral_contribution, n=n)
 
 
+    def _enforce_ratio(self, goal_ratio, supx, infx, supy, infy):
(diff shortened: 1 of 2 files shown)


fAndreuzzi:
In 296a8b6 I made a small change to remove the arrow of an axis when the point `1.0` of that axis is not visible. The image attached provides a better explanation.

image

PR #183, opened 2021-08-02
Hankel DMD add, HODMD fix

### LINKED ISSUE #132: HODMD doesn't agree with the reference
By looking at the code, the current HODMD seems to be a "Hankel-DMD"-- first stack the staggered data, next perform DMD on the new data. 

But I found the [code](https://github.com/LeClaincheVega/HODMD/blob/master/DMDd_SIADS.m) and paper for the original HODMD are different than the pydmd. Their first step is to reduce the data.

I think they are not equivalent.

ndem0:
Fix #132

### DIFF
--- a/pydmd/hankeldmd.py
+++ b/pydmd/hankeldmd.py
@@ -1,0 +1,188 @@
+"""
+Derived module from dmdbase.py for hankel dmd.
+
+Reference:
+- H. Arbabi, I. Mezic, Ergodic theory, dynamic mode decomposition, and
+computation of spectral properties of the Koopman operator. SIAM Journal on
+Applied Dynamical Systems, 2017, 16.4: 2096-2126.
+"""
+import numpy as np
+
+from .dmdbase import DMDBase, DMDTimeDict
+from .utils import compute_tlsq
+
+
+class HankelDMD(DMDBase):
+    """
+    Higher Order Dynamic Mode Decomposition
+
+    :param svd_rank: the rank for the truncation; If 0, the method computes the
+        optimal rank and uses it for truncation; if positive interger, the
+        method uses the argument for the truncation; if float between 0 and 1,
+        the rank is the number of the biggest singular values that are needed
+        to reach the 'energy' specified by `svd_rank`; if -1, the method does
+        not compute truncation.
+    :type svd_rank: int or float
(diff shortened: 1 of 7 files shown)


pydmd/hankeldmd.py
```diff
@@ -0,0 +1,187 @@
...
+"""
+import numpy as np
+
+from .dmdbase import DMDBase, DMDTimeDict
+from .utils import compute_tlsq
+
+
+class HankelDMD(DMDBase):
+    """
+    Higher Order Dynamic Mode Decomposition
```
mtezzele, line 17: Old name

pydmd/utils.py
```diff
@@ -26,3 +26,44 @@ def compute_tlsq(X, Y, tlsq_rank):
...
+        the optimal rank and uses it for truncation; if positive interger,
+        the method uses the argument for the truncation; if float between 0
+        and 1, the rank is the number of the biggest singular values that
+        are needed to reach the 'energy' specified by `svd_rank`; if -1,
+        the method does not compute truncation. Default is 0.
+    :type svd_rank: int or float
+    :return: the truncated left-singular vectors matrix, the truncated
+        singular values array, the truncated right-singular vectors matrix.
+    :rtype: numpy.ndarray, numpy.ndarray, numpy.ndarray
+    References:
```
mtezzele, line 44: Maybe a return before is needed

tests/test_hodmd.py
```diff
@@ -312,6 +290,22 @@ def test_rec_method_mean(self):
         assert (dmd.reconstructed_data.T[2] == np.mean(dmd.reconstructions_of_timeindex(2), axis=0).T).all()
 
     def test_rec_method_weighted(self):
-        dmd = HODMD(d=2, reconstruction_method=[10,20])
+        dmd = HODMD(d=2, svd_rank_extra=-1,reconstruction_method=[10,20])
         dmd.fit(X=sample_data)
         assert (dmd.reconstructed_data.T[4] == np.average(dmd.reconstructions_of_timeindex(4), axis=0, weights=[10,20]).T).all()
+    """
```
mtezzele, line 296: should this remain?

ndem0: I would keep it for future: as soon as I check the dynamics of this version, I'll update the (hardcoded) expected dynamics and restore the test!

PR #185, opened 2021-09-05
HankelDMD uses an auxiliary DMD instance to preserve timesteps

fAndreuzzi:
In this PR I implemented the following features/fixes in the classes `HODMD, HankelDMD`:
- show warning and use `svd_rank_extra=-1` when `svd_rank_extra=0` and the system is scalar function (one row);
- return `self` in `HODMD.fit` (as we do in all the other implementations);
- replace `np.outer` to repeat the eigenvalues `n` times with the NumPy method `np.repeat` (which is clearer in my opinion);
- `HankelDMD` now contains an auxiliary instance of `DMD` which operates on Hankel pseudo-matrix:
    + This kind of approach is preferable in order to allow `dmd_timesteps` to work as intended, for instance it fixes the error at https://github.com/mathLab/PyDMD/blob/6de9464311a5fde4843f6fa55b6b324eead57a2c/tutorials/tutorial-6-hodmd.py#L22-L30
    + Some work to update the values in `HankelDMD._sub_dmd.dmd_time` after a change to `HankelDMD.dmd_time`was needed (check `HankelDMD. _update_sub_dmd_time` and `HankelDMD._hankel_first_occurrence`);
    + The properties `modes`, `amplitudes`, `operator` and `svd_rank` of `HankelDMD` return the corresponding values of `HankelDMD._sub_dmd`;
- I implemented new tests to verify that the changes integrated in this PR behave as expected.
- I ran the formatter on the files which I worked on in this PR.

### DIFF
--- a/pydmd/hankeldmd.py
+++ b/pydmd/hankeldmd.py
@@ -10,6 +10,7 @@ import numpy as np
 
 from .dmdbase import DMDBase, DMDTimeDict
 from .utils import compute_tlsq
+from .dmd import DMD
 
 
 class HankelDMD(DMDBase):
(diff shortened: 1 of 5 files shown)


pydmd/hodmd.py
```diff
@@ -116,9 +133,20 @@ def fit(self, X):
         """
         org_snp, snapshots_shape = self._col_major_2darray(X)
 
-        self.U_extra, _, _ = compute_svd(org_snp, self.svd_rank_extra)
+        if org_snp.shape[0] == 1:
+            self.U_extra, _, _ = compute_svd(org_snp, -1)
+            warnings.warn(
+                "The parameter 'svd_rank_extra={}' has been ignored because the given system is a scalar function".format(
```
ndem0, line 139: Line exceeding max length

pydmd/hankeldmd.py
```diff
@@ -54,26 +55,75 @@ class HankelDMD(DMDBase):
...
+            rescale_mode=rescale_mode,
+            forward_backward=forward_backward,
+            sorted_eigs=sorted_eigs,
+        )
+
     @property
     def d(self):
         return self._d
 
+    def _hankel_first_occurrence(self, time, is_original_t0=False):
```
ndem0, line 109: Docstring is missing

pydmd/hankeldmd.py
```diff
@@ -54,26 +55,75 @@ class HankelDMD(DMDBase):
...
+        if is_original_t0:
+            return 0
+        else:
+            return max(
+                0,
+                (time - self.original_time["t0"]) // self.dmd_time["dt"]
+                - (self.original_time["t0"] + self.d - 1),
+            )
+
+    def _update_sub_dmd_time(self):
```
ndem0, line 119: Docstring is missing

fAndreuzzi:
@ndem0 In 0863a48 I resolved the problems which you raised in your comments.

PR #188, opened 2021-09-23
Fix e-mail issue

fAndreuzzi:
In this commit I tried to fix the bug notified by a user via e-mail.

> I noted that in partial dynamics code (within mrdmd module has timestep (1600) is probably hardcoded. Possibly, the value 1600 comes from the tutorial 3 (MrDMD) time steps. This results in error while working on other possible time steps.

Even though the problem looks fixed and the script runs without errors I notified some problems while running the new test which I introuced in `test_mrdmd.py`:

```python
    def test_consistency2(self):
        import sys
        import numpy
        numpy.set_printoptions(threshold=sys.maxsize)


        mrdmd = MrDMD(DMD(), max_level=5, max_cycles=1)
        mrdmd.fit(X=create_data(t_size=1400))


        np.testing.assert_array_almost_equal(mrdmd.reconstructed_data, mrdmd.modes @ mrdmd.dynamics)
```

This is the same kind of test of `test_consistency`, but uses a different number of time instants than 1600. When running the test, I get the error:

```
 AssertionError: 
       Arrays are not almost equal to 6 decimals
```

Since I do not know if my change breaks something in mrDMD I'm marking this PR as a draft to be reviewed.

PS: I could not reproduce the error reported by the same user with `plot_eigs`

### DIFF
--- a/pydmd/mrdmd.py
+++ b/pydmd/mrdmd.py
@@ -291,9 +291,9 @@ class MrDMD(DMDBase):
         dynamics = []
         for i, leaf in enumerate(leaves):
             d = self.dmd_tree[level, leaf].dynamics
-            base = np.zeros(shape=(d.shape[0], 1600), dtype=d.dtype) #TODO
+            base = np.zeros(shape=(d.shape[0], self._snapshots.shape[1]), dtype=d.dtype)
             time_interval = self.partial_time_interval(level, leaf)
-            base[:, int(time_interval['t0']):int(time_interval['tend'])] = d
+            base[:, int(time_interval['t0']):int(time_interval['t0'])+d.shape[1]] = d
             dynamics.append(base)
 
         dynamics = np.vstack(dynamics)
--- a/pydmd/mrdmd.py
+++ b/pydmd/mrdmd.py
@@ -12,6 +12,7 @@ import numpy as np
 import scipy.linalg
 import matplotlib.pyplot as plt
 from copy import deepcopy
+from scipy.linalg import block_diag
 
 from .dmdbase import DMDBase, DMDTimeDict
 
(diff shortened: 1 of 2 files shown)


fAndreuzzi:
I found out that using `scipy.linalg.block_diag` ([source](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.block_diag.html)) to compute the matrix of partial dynamics fixes the error in the test which I introduced in 5487f35. However I had to change two tests a little bit (`test_partial_dynamics2`, `test_partial_reconstructed2`) to make them pass without errors, could someone confirm that the new check is reasonable @ndem0.

pydmd/mrdmd.py
```diff
@@ -287,16 +286,9 @@ def partial_dynamics(self, level, node=None):
...
-        dynamics = []
-        for i, leaf in enumerate(leaves):
-            d = self.dmd_tree[level, leaf].dynamics
-            base = np.zeros(shape=(d.shape[0], 1600), dtype=d.dtype) #TODO
-            time_interval = self.partial_time_interval(level, leaf)
-            base[:, int(time_interval['t0']):int(time_interval['tend'])] = d
-            dynamics.append(base)
-
-        dynamics = np.vstack(dynamics)
+        print(leaves)
```
ndem0, line 289: remove this line

ndem0:
> I found out that using `scipy.linalg.block_diag` ([source](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.block_diag.html)) to compute the matrix of partial dynamics fixes the error in the test which I introduced in [5487f35](https://github.com/mathLab/PyDMD/commit/5487f355f2f2b2b242e32676e967b33c151ee85e). However I had to change two tests a little bit (`test_partial_dynamics2`, `test_partial_reconstructed2`) to make them pass without errors, could someone confirm that the new check is reasonable @ndem0 ?

It seems to me that the implementations are basically the same, but number of rows in the returned array is different. To avoid confusion, can we specify the dimensions of the dynamics?

fAndreuzzi:
@ndem0 Could you please explain what do you mean with
> can we specify the dimensions of the dynamics?

ndem0:
> @ndem0 Could you please explain what you mean with
> 
> > can we specify the dimensions of the dynamics?

Dimensions (n. rows, n. columns) of the returned array calling the `dynamics` property

Issue #184, opened 2021-08-23
Regularization Enhancement?


rajarshiroychowdhury:
Was wondering if an enhancement can be added when the input matrix is ill-conditioned. The warning saying the condition number is high certainly helps but I was wondering if it's possible to implement the regularization as described on page 8 of the Erichson et al paper attached below
[Random-DMD_Erichson.pdf](https://github.com/mathLab/PyDMD/files/7029255/Random-DMD_Erichson.pdf)
. I think this will at least give a way to handle the case of an ill-conditioned input matrix.

mtezzele:
Dear @rajarshiroychowdhury, regularization is for sure a possible way to handle such problem. It is possible we will add this feature in the future. In the near term we are going to integrate new DMD versions. If you want to contribute on the regularization we would be very glad.

rajarshiroychowdhury:
Thanks for the response @mtezzele! Appreciate it! I probably would have to dig through the code at some point and get this done, but I am sure the main developers can implement this pretty quickly. It seems like just adding another term with a parameter that the user might be able to specify while calling let's say hodmd. Again, I understand and appreciate the fact that possibly integrating new DMD variations is the near-term goal, but I just thought this should be relatively easy and straightforward for the developers than someone else who might need to dig into the code to figure out where exactly this should be added. Regardless, thank you!

PR #241, opened 2022-01-14
Regularization

azzeddinetiba:
Added Tikhonov regularization to deal with ill-conditioned input matrices, as mentioned in #184 and #107 .
The regularization is implemented as a filter on the SVD reconstruction of **X** , the filter being sigma^2 / (sigma^2 + lambda), with lambda = ```tikhonov```*||**X**||.

1. Tests passed on Ubuntu 18.04, Python 3.8
2. I added a "tutorial11", it is not a tutorial per se, but it shows the ffect of the regularization on fluid simulation data. The .npy data file is just above 100Mo though, I can't push it. It shows this effect on DMD modes only, I did see the effect on instabilites of the DMD prediction on other cases.
3. Should the bad conditioning warning (92b33ee3e87b1483bc36ef0bd70664393cff1f6e) be ommited in case regularization is wanted ?


ndem0:
Dear @azzeddinetiba, first of all many thanks for the nice new feature and nice tutorial!
Before merging it, I have just few questions/requests:
- maybe we can use `tikhonov_regularization` instead of just `tikhonov` for the attribute name? In this way also people not really expert in the field can easily exploit such attribute for ill-conditioned data (with only `tikhonov` it looks quite criptic to me);
- for being more "pythonic", probably to disable the regularization it is better to have the `tikhonov` parameter to `None` instad of 0
- Can you solve the static issues that Codacy have spotted? https://app.codacy.com/gh/mathLab/PyDMD/pullRequest?prid=8712068
- Regarding the bad condition warning, we have planned to better organize the output of PyDMD, e.g with the logging module. So I would say we can skip manual print for the moment.

azzeddinetiba:
Thank you @ndem0 for responding
 
It should be all good now.

ndem0:
Awesome!
Thanks again, once the tests will be concluded I'll merge your additions.

pydmd/cdmd.py
```diff
@@ -37,13 +37,19 @@ class CDMDOperator(DMDOperator):
         magnitude if `sorted_eigs='abs'`, by real part (and then by imaginary
         part to break ties) if `sorted_eigs='real'`. Default: False.
     :type sorted_eigs: {'real', 'abs'} or False
+    :param tikhonov: tikhonov parameter for regularization
+        If None, no regularization is applied, if float, it is used as the
+        lambda tikhonov parameter
```
fAndreuzzi, line 42:
```suggestion
    :param tikhonov: Tikhonov parameter for the regularization.
        If `None`, no regularization is applied, if `float`, it is used as the
        :math:`\lambda` tikhonov parameter.
```

azzeddinetiba:
I also forgot to update the tutorial to tkhonov = None instead of 0. Sorry about that. I'll modify it asap.

fAndreuzzi:
> I also forgot to update the tutorial to tkhonov = None instead of 0. Sorry about that. I'll modify it asap.

Thanks!

azzeddinetiba:
Docstring and tutorial updated.

azzeddinetiba:
Last commit for a docstring occurence I missed.

fAndreuzzi:
Thank you, I was about to write about the missing update to the docstring in `cdmd.py`. 

One more question, can you explain why you decided not to include the new parameter `tikhonov_regularization` in all the DMD subclasses? I see that you modified only `cdmd.py`, `dmd.py`, `optdmd.py`, `dmdc.py`.

azzeddinetiba:
@fAndreuzzi Hello back. I used this regularization previously on DMD and DMD with control only. I am not familiar with all the other variations. Although I supect they all use the same SVD-based dmd operator, I've thought to let it to the according authors to decide whether to include the regularization or not, in case these variations treat this problem using some other particular strategies.

PR #250, opened 2022-02-17
[ModesTuner] Introduce parameter `nullify_amplitudes`

fAndreuzzi:
Changelog:
- Introduced the parameter `nullify_amplitudes` (defaults to `False`) which allows setting to zero the amplitudes corresponding to DMD modes to be cut in `select_modes()` or `ModesTuner.select()`;
- Improved test coverage (some lines were uncovered since my last PR);
- `ModesTuner.select()/stabilize()` now return `self` (i.e. the calling instance of `ModesTuner`) in order to allow chaining multiple operations with less code.

For instance, this is now valid:
```python
new_dmd1, new_dmd2 = (
        ModesTuner([dmd1, dmd2])
        .select("integral_contribution", n=5)
        .select(
            "module_threshold",
            low_threshold=1 - 1.0e-5,
            up_threshold=1 + 1.0e-2,
        )
        .stabilize(inner_radius=1 - 1.0e-3)
        .get()
    )
```

### DIFF
--- a/pydmd/__init__.py
+++ b/pydmd/__init__.py
@@ -17,3 +17,4 @@ from .dmdc import DMDc
 from .optdmd import OptDMD
 from .spdmd import SpDMD
 from .paramdmd import ParametricDMD
+from .dmd_modes_tuner import ModesTuner
--- a/pydmd/cdmd.py
+++ b/pydmd/cdmd.py
@@ -154,14 +154,14 @@ class CDMD(DMDBase):
         self._opt = opt
         self._compression_matrix = compression_matrix
 
-        self._modes_activation_bitmask = None
-
         self._Atilde = CDMDOperator(svd_rank=svd_rank,
                                     rescale_mode=rescale_mode,
                                     forward_backward=forward_backward,
                                     sorted_eigs=sorted_eigs,
                                     tikhonov_regularization=
                                     tikhonov_regularization)
+
+        self._modes_activation_bitmask_proxy = None
 
     @property
     def compression_matrix(self):
--- a/pydmd/dmd_modes_tuner.py
+++ b/pydmd/dmd_modes_tuner.py
@@ -66,27 +66,10 @@ def select_modes(
     all_indexes = set(np.arange(len(dmd.eigs)))
     cut_indexes = np.array(list(all_indexes - set(selected_indexes)))
 
-    if nullify_amplitudes:
-        dmd._b[cut_indexes] = 0
-    else:
-        dmd.operator._eigenvalues = dmd.operator._eigenvalues[selected_indexes]
-        dmd.operator._Lambda = dmd.operator._Lambda[selected_indexes]
-
-        dmd.operator._eigenvectors = dmd.operator._eigenvectors[
-            :, selected_indexes
-        ]
-        dmd.operator._modes = dmd.operator._modes[:, selected_indexes]
-
-        # TODO: should improve this [code repetition]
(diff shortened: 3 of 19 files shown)


fAndreuzzi:
In 789b21a I introduced some changes which I would like to discuss with the team since they are quite "intensive" in some sense. This is the result of several experiments I did with a few different approaches, and I think it's the cleanest and easiest to maintain.

As discussed with @ndem0 the problem was related to the fact that double indexing in NumPy doesn't produce "views" on the actual array: instead they return a full copy of the selection, which means that we cannot assign to the original array (which usually is an array of amplitudes or eigenvalues). This is interesting, for instance, for the following use case of `dmd_modes_tuner.py`: a first indexing is performed using the *activation bitmask* (#245), and then another one in `dmd_modes_selector.stabilize_modes()`. This second selection should also modify the elements in the selected sub-array, but as I mentioned this is not possible.

For this reason I introduced the class `dmdbase.ActivationBitmaskProxy`, which holds temporary copied of `eigs`, `amplitudes`, `modes` induced by a particular value of the activation bitmask. Modification on these copied is allowed, therefore these are what is now returned by `DMDBase.amplitudes/eigs/modes`. Most importantly, modification on these copies are reflected on the original values of `eigs`, `amplitudes`, `modes` (the logic is managed by `ActivationBitmaskProxy`). 

In case the approach is approved I'll document and test the new class (the 7 new problems on Codacy are all related with the lack of docstrings).

ndem0:
The new class is fine, it is encapsulated so the complexity for the user still be the same.
I have some doubts about the example you show in docstring:
```
            >>> # this is an error
            >>> dmd.modes_activation_bitmask[[1,2]] = False
            ValueError: assignment destination is read-only
            >>> tmp = np.array(dmd.modes_activation_bitmask)
            >>> tmp[[1,2]] = False
            >>> dmd.modes_activation_bitmask = tmp
```
Is it really necessary a 2D array for indexing?
To be honest, I did not get the purpose of closing the writing permission on the bitmask array, since you can set such attribute by passing any object. Potentially you can meet the same bugs as before, but requiring the triple number of lines of code. Did I lose something?

fAndreuzzi:
Hi @ndem0,

> To be honest, I did not get the purpose of closing the writing permission on the bitmask array, since you can set such attribute by passing any object. Potentially you can meet the same bugs as before, but requiring the triple number of lines of code. Did I lose something?

The reason for which I decided to enforce this is this call which occurs in the setter for the property `DMDBase.modes_activation_bitmask`:
https://github.com/mathLab/PyDMD/blob/789b21ae4f7e16662665c02d91185c00e5d46675/pydmd/dmdbase.py#L449

If for instance we do this:
```python
dmd.modes_activation_bitmask[[1,2]] = False
```
Then the setter is not called, and the bitmask changes without any notification to the proxy. This breaks some invariants between the enclosing `DMDBase` instance and the proxy, since the proxy does some critical modifications in `change_bitmask`.

> Is it really necessary a 2D array for indexing?

I'm not sure whether I got your question properly, but it's not a 2D array as far as I know. Here I'm just setting the second and this element of `tmp` to `False` using a list of indexes:
```python
>>> tmp[[1,2]] = False
```
Am I missing something?

ndem0:
> I'm not sure whether I got your question properly, but it's not a 2D array as far as I know. Here I'm just setting the second and this element of tmp to False using a list of indexes:

Yes, you're right, I've read too fast the code. It's just a list passed to the `__getitem__` method! Great, I merge everything!

PR #293, opened 2022-10-13
Added Randomized DMD

sichinaga:
Here is the `rdmd.py` file which implements the `RDMD` class for Randomized DMD. This module builds off of the Compressed DMD module since the methods are very similar.


pydmd/rdmd.py
```diff
@@ -0,0 +1,89 @@
+"""
+Derived module from cdmd.py for Randomized DMD
+References:
+- N. Benjamin Erichson, Lionel Mathelin, J. Nathan Kutz, and Steven L. Brunton. Randomized
+dynamic mode decomposition. SIAM Journal on Applied Dynamical Systems, 18, 2019.
+"""
+
+import numpy as np
+from .cdmd import CDMD, CDMDOperator
```
mtezzele, line 9: unused CDMDOperator

pydmd/rdmd.py
```diff
@@ -0,0 +1,89 @@
...
+from .utils import compute_tlsq
+
+def rqb(X, k, p, q):
```
mtezzele, line 12: I would use more verbose variable/method names. For example `randomized_qb` instead of `rqb`. Same applies for k, p, q. Just for an easier maintenance in the future.

pydmd/rdmd.py
```diff
@@ -0,0 +1,89 @@
...
+    :return: a near-optimal orthonormal basis Q for the range of the matrix X, 
+        and the projection B of the matrix X onto the low-dimensional space Q.
+    :rtype: numpy.ndarray, numpy.ndarray
+
+    References:
+    N. Benjamin Erichson, Lionel Mathelin, J. Nathan Kutz, and Steven L. Brunton. 
+    Randomized dynamic mode decomposition. SIAM Journal on Applied Dynamical Systems, 18, 2019.
+    """
+    n, m = X.shape
+    l = k + p # slight oversampling
```
mtezzele, line 30: Please put the comments above the code lines

pydmd/rdmd.py
```diff
@@ -0,0 +1,89 @@
...
+    Omega = np.random.randn(m, l) # generate random test matrix 
+    Y = X @ Omega # compute sampling matrix 
+    for j in range(q): # power iterations 
+        Q, _ = np.linalg.qr(Y)
```
mtezzele, line 34: Instead of `_` you can use the indexing on the rhs:
`Q = np.linalg.qr(Y)[0]`

pydmd/rdmd.py
```diff
@@ -0,0 +1,89 @@
...
+    return Q, B
+
+class RDMD(CDMD):
+    """
+    Randomized Dynamic Mode Decomposition
+    :param int p: Randomized QB decomposition oversampling parameter 
+    :param int q: Randomized QB decomposition power iterations 
+    """
+    
+    def __init__(self, svd_rank, p=10, q=2, tlsq_rank=0, opt=False, rescale_mode=None,
```
mtezzele, line 48: more verbose args names

mtezzele (commented):
Thank you for the contribution. The readme part is perfect. 
For the `RDMD` please have a look at the [codacy report](https://app.codacy.com/gh/mathLab/PyDMD/pullRequest?prid=10354462), there are some unused variables, lines exceeding the 80 chars limit, and trailing spaces. I meda some suggestions regarding the variable names. You can keep the matrices as single capital letters since the meaning is straightforward, for the rest please consider more verbose options.

mtezzele:
@ndem0 what do you think about the `rqb` method? Should we put it inside the class as static method?

sichinaga:
Hi @mtezzele and @ndem0! Thanks for all of the great feedback! I believe all of the issues that have been brought up are now resolved, with the exception of what looks to be a tiny amount of codacy isses that aren't major. My apologies for the messy code. I'll work on making sure my code is extra readable in future commits. Let me know if there are any more edits that you would suggest! 

Also, one tricky thing about this method is that the randomized QB decomposition requires a "target rank," which is basically the `svd_rank` of the model. I currently have it so that `svd_rank` must be provided by the user to avoid essentially copying and pasting the code from `compute_svd` where `rank` is computed from a variety of possible `svd_rank` inputs. But ideally, `svd_rank` should have a default value of `0` like in all other DMD classes. 

Basically, should the  `rank` computation code just go into `RDMD`? Or would it be better to split `compute_svd` into two methods, where one part computes the rank-truncated SVD while the other part computes the rank?

mtezzele:
@sichinaga thank you for the edits! Now the last important piece is to write some tests. I suggest you start with a copy/paste of the CDMD tests (`tests/test_cdmd.py`) and change the names. The majority of the tests are trivial and just check that the arguments are correctly read. Regarding the rank computation I rely on @ndem0 and @fAndreuzzi suggestions.

pydmd/rdmd.py
```diff
@@ -0,0 +1,126 @@
...
+    m = X.shape[1]
+
+    # Generate random test matrix (with slight oversampling)
+    Omega = np.random.randn(m, target_rank + oversampling)
+
+    # Compute sampling matrix
+    Y = X.dot(Omega)
+
+    # Perform power iterations
+    for j in range(power_iters):
```
fAndreuzzi, line 44: Apparently `j` is not used anywhere in the loop, so you could just replace it with `_` (this fixes a Codacy warning).

```suggestion
    for _ in range(power_iters):
```

pydmd/rdmd.py
```diff
@@ -0,0 +1,126 @@
...
+N. Benjamin Erichson, Lionel Mathelin, J. Nathan Kutz, Steven L. Brunton.
+Randomized dynamic mode decomposition. SIAM Journal on Applied Dynamical
+Systems, 18, 2019.
+"""
+
+import numpy as np
+from .cdmd import CDMD
+from .utils import compute_tlsq
+
+def randomized_qb(X, target_rank, oversampling, power_iters):
```
fAndreuzzi, line 14: For me, I would put this function as a private instance method into the class `RDMD` for two main reasons:
- All the parameters for `randomized_qb` are actually instance variables of `RDMD`
- We don't really need to expose this method to the user, and we don't need to have Sphinx generate documentation for us on this function.

@ndem0 @mtezzele what do you think?

fAndreuzzi:
Hi @sichinaga, thank you for your PR! left some minor comments on your code.

I like the idea of splitting `compute_svd` in two methods to have separate computation of the SVD rank such that we can re-use it in other places, maybe the name for the new function could be something like `compute_n_svd_modes`? The function would stay together with `compute_svd` in `utily.py`

@ndem0 do you agree?

ndem0:
Dear @sichinaga, thanks for the effort! The code is pretty fine, but I still have a couple of questions:
- probably, the cleanest way to implement RDMD would be just re-implement the `__init__` with the new parameters and encapsulate the `randomized_qb` in the `_compress_snapshots` method from CDMD. In this way, I think we do not need to rewrite the `fit` method and everything becomes easier to maintain in the future (less code is typically better). Can you try such refactoring after having implemented some tests, as suggested by @mtezzele?
- Actually, I don't see why splitting the method `compute_svd` in 2 different ones @fAndreuzzi. Where I am lost?

sichinaga:
@fAndreuzzi @ndem0 Thank you both so much for the feedback! It's been very helpful. I also apologize for my brief PyDMD hiatus (a lot of fellowship stuff have been due lately), but I am back now! 

Anyway to summarize the more recent changes, @fAndreuzzi @ndem0, I implemented your suggestions by placing the `randomized_qb` code into the private `_compress_snapshots` method. Basically the randomized qb decomposition is now performed in the `_compress_snapshots` method, so now `RDMD` uses the `fit` method from `CDMD`.

Additionally, on the topic of splitting `compute_svd`, @fAndreuzzi, could you potentially elaborate more on what `compute_n_svd_modes` would do as opposed to `compute_svd`? I just want to proceed with caution since `compute_svd` is kind of ubiquitous throughout PyDMD and I want to fiddle with it as little as possible. I sort of envisioned that the rank computation could be performed in some method called `compute_rank` and its output could get passed to what is currently `compute_svd`, but did you have something else in mind? 

On that note, @mtezzele, once the rank-computation bit is squared away, I'll be sure to implement the necessary tests!

sichinaga:
Update: Added a skeleton for the `RDMD` tests, which is basically all of the relevant tests for `CDMD` but renamed and/or commented out because the svd rank functionality needs to be implemented first.

ndem0:
Don't worry so much about the delay, and thanks for the great job you already did!
The code is quite fine, once we fix the point around splitting `compute_svd` and the tests we gonna merge the new module!

Regarding the open questions:

>  I sort of envisioned that the rank computation could be performed in some method called compute_rank and its output could get passed to what is currently compute_svd, but did you have something else in mind?

Yes, I agree, it looks like the easiest way. You can also (temporarily) add the two new methods (`compute_svd` and `compute_rank`) inside the module `rdmd.py`, removing the import from utils, such that you can initially test your implementation only within RDMD class. Once you're happy with the code, we gonna try to use the new functions in all the classes, but at least at the beginning, you have to fight with fewer errors!

sichinaga:
Thanks @ndem0! The code for `compute_rank` and all of the tests have now been added. I've also fixed as many Codacy issues as possible! Thanks for all of your help, and let me know if there's anything else I can add/fix regarding this PR! :)

PR #295, opened 2022-11-10
Adding HAVOK to PyDMD

sichinaga:
Hello again to everyone on the PyDMD team!

I am here to propose the next potential addition to the PyDMD family: the Hankel Alternative View of Koopman (HAVOK) analysis! In case anyone is interested, the article deriving and describing the method can be found here: https://www.nature.com/articles/s41467-017-00030-8

HAVOK is a method specifically tailored for modeling one-dimensional time-series datasets that exhibit chaos. In short, the algorithm can be broken down into the following steps:

1) Compute a hankel matrix $H$ from the input time-series.
2) Compute the SVD of the hankel matrix $H = U\Sigma V^*$.
3) Apply DMD to $V^T$ and obtain the continuous-time DMD operator $A_c$ in $\dot{V}^T = A_c V^T$.
4) Define $A$ as the matrix $A_c$ but excluing the last row and column. Define $B$ as the last column of $A_c$, but excluding the last entry.
5) We may then model our data via the relationship $\dot{v} = Av + Bv_r$, where $V = [v_1 \quad \dots \quad v_r]$ and $v = [v_1 \quad \dots \quad v_{r-1}]^T$.

Given the structure of the HAVOK analysis, I have made the following implementation choices:
- The `HAVOK` class is a child of the `HankelDMD` class due to their shared usage of hankel matrices. Note that the key differences between the two methods are (1) HAVOK applies DMD to the $V$ matrix from the SVD of $H$ while Hankel DMD applies DMD to $H$, and (2) HAVOK utilizes a forcing term in order to reconstruct data, while Hankel DMD does not.
- In order to utilize currently-existing DMD modules, I compute the continuous-time operator $A_c$ by first using DMD to compute the discrete-time operator $A_d$ in $V_2^T = A_d V_1^T$. I then utilize the approximation $A_c \approx (A_d - I) / \Delta t$. There are most certainly better/more accurate methods of computing $A_c$, but I figured this rough approximation should suffice for now.
- I'm currently using `scipy.signal` in order to compute the HAVOK reconstructions (since this is how reconstructions are computed in the original HAVOK paper, and the code is quite concise), but I could also envision this being implemented via the `DMDc` module somehow... I'm happy to keep the reconstruction code as is, but if reusing code from `DMDc` has the potential to be stylistically better, I'd be happy to give the implementation a shot!
- One other thing, I've added a `tikhonov_regularization` input parameter to `HankelDMD` so that it may now be used by both `HankelDMD` and `HAVOK`. Let me know if this parameter was omitted from `HankelDMD` for any particular reason so that I can restore `HankelDMD` to its original form! :)

Thank you all again for your diligence and support with these new modules. 😄 

I'm also open to any feedback, comments, questions, or concerns that may arise with this module. :)

### DIFF
--- a/pydmd/havok.py
+++ b/pydmd/havok.py
@@ -139,9 +139,8 @@ class HAVOK(HankelDMD):
             np.diag(self._svd_amps[:-1]),
             reconstructed_v.conj().T
         ])
-        reconstructed_x = self._dehankel(reconstructed_hankel_matrix)
 
-        return reconstructed_x
+        return self._dehankel(reconstructed_hankel_matrix)
 
     def fit(self, x, dt):
         """
@@ -152,7 +151,7 @@ class HAVOK(HankelDMD):
         self._snapshots = self._snapshots.squeeze()
 
         # Check that input data is a 1D time-series
-        if np.ndim(self._snapshots) != 1:
+        if self._snapshots.ndim > 1:
             raise ValueError("Input data must be a 1-D time series.")
 
         # Get number of data points
@@ -165,9 +164,7 @@ Expected at least d."""
             raise ValueError(msg.format(self._d))
 
         # Compute hankel matrix for the input data
-        self._hankel_matrix = self._pseudo_hankel_matrix(
-            self._snapshots.reshape(1, -1)
-        )
+        self._hankel_matrix = self._pseudo_hankel_matrix(self._snapshots[None])
 
         # Take SVD of the hankel matrix
         # Save the resulting U, s, and V for future reconstructions
@@ -192,7 +189,7 @@ Expected at least d."""
 
         # Save A matrix and B vector
         self._A = regression_model_continuous[:-1, :-1]
-        self._B = regression_model_continuous[:-1, -1].reshape(-1, 1)
+        self._B = regression_model_continuous[:-1, -1, None]
 
         # Set timesteps
         self._set_initial_time_dictionary(
--- a/pydmd/havok.py
+++ b/pydmd/havok.py
@@ -35,7 +35,7 @@ class HAVOK(HankelDMD):
             opt=opt,
             rescale_mode=rescale_mode,
             forward_backward=forward_backward,
-            sorted_eigs=sorted_eigs, 
+            sorted_eigs=sorted_eigs,
             tikhonov_regularization=tikhonov_regularization,
             d=d,
         )
@@ -52,6 +52,8 @@ class HAVOK(HankelDMD):
         :return: matrix containing the linear time-delay embeddings.
         :rtype: numpy.ndarray
         """
+        if self._embeddings is None:
+            raise RuntimeError("fit() not called")
         return self._embeddings[:,:-1]
 
     @property
@@ -62,6 +64,8 @@ class HAVOK(HankelDMD):
         :return: array containing the chaotic forcing term.
         :rtype: numpy.ndarray
         """
+        if self._embeddings is None:
+            raise RuntimeError("fit() not called")
         return self._embeddings[:,-1]
 
     @property
@@ -73,6 +77,8 @@ class HAVOK(HankelDMD):
         :return: linear dynamics matrix A.
         :rtype: numpy.ndarray
         """
+        if self._A is None:
+            raise RuntimeError("fit() not called")
         return self._A
 
     @property
@@ -84,6 +90,8 @@ class HAVOK(HankelDMD):
         :return: forcing dynamics vector B.
         :rtype: numpy.ndarray
         """
+        if self._B is None:
+            raise RuntimeError("fit() not called")
         return self._B
 
     def _dehankel(self, X):
@@ -100,7 +108,7 @@ class HAVOK(HankelDMD):
 
     @property
     def reconstructed_data(self):
-        # Build a continuous-time system of the following form, where 
+        # Build a continuous-time system of the following form, where
         # x, u, and y represent the states, inputs, and outputs respectively.
         #
         # dx/dt = Ax + Bu
@@ -182,7 +190,7 @@ Expected at least d."""
             regression_model_discrete - np.eye(self._r)
         ) / dt
 
-        # Save A matrix and B vector 
+        # Save A matrix and B vector
         self._A = regression_model_continuous[:-1, :-1]
         self._B = regression_model_continuous[:-1, -1].reshape(-1, 1)
 
--- a/pydmd/havok.py
+++ b/pydmd/havok.py
@@ -1,0 +1,194 @@
+"""
+Derived module from hankeldmd.py for HAVOK.
+
+Reference:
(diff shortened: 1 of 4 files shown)


pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def linear_embeddings(self):
+        """
+        Get the time-delay embeddings of the data that are governed by linear
+        dynamics. Emeddings are stored as columns of the returned matrix.
+
+        :return: matrix containing the linear time-delay embeddings.
+        :rtype: numpy.ndarray
+        """
+        return self._embeddings[:,:-1]
```
fAndreuzzi, line 55: This will trigger a weird error if `fit()` is not called

```suggestion
        if self._embeddings is None:
                raise RuntimeError('fit() not called')
        return self._embeddings[:,:-1]
```

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @property
+    def forcing_input(self):
+        """
+        Get the time-delay embedding that forces the linear embeddings.
+
+        :return: array containing the chaotic forcing term.
+        :rtype: numpy.ndarray
+        """
+        return self._embeddings[:,-1]
```
fAndreuzzi, line 65: Same as `linear_embeddings`

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def A(self):
+        """
+        Get the matrix A in the relationship dV/dt = AV + Bu, where V denotes
+        the linear embeddings and u denotes the forcing input.
+
+        :return: linear dynamics matrix A.
+        :rtype: numpy.ndarray
+        """
+        return self._A
```
fAndreuzzi, line 76: I believe it's good to throw an exception here too, like for `linear_embeddings`

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def B(self):
+        """
+        Get the vector B in the relationship dV/dt = AV + Bu, where V denotes
+        the linear embeddings and u denotes the forcing input.
+
+        :return: forcing dynamics vector B.
+        :rtype: numpy.ndarray
+        """
+        return self._B
```
fAndreuzzi, line 87: Same as `A` above

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+        # the hankel matrix with the reconstructed V matrix and then recovering
+        # a 1-D time-series from the computed hankel matrix.
+        reconstructed_hankel_matrix = np.linalg.multi_dot([
+            self._svd_modes[:, :-1],
+            np.diag(self._svd_amps[:-1]),
+            reconstructed_v.conj().T
+        ])
+        reconstructed_x = self._dehankel(reconstructed_hankel_matrix)
+
+        return reconstructed_x
```
fAndreuzzi, line 136: I propose to merge these just to make it more compact

```suggestion
        return self._dehankel(reconstructed_hankel_matrix)
```

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+    def fit(self, x, dt):
+        """
+        Perform HAVOK analysis on 1-D time-series data x given the size of
+        the time step dt separating the observations in x.
+        """
+        self._snapshots, self._snapshots_shape = self._col_major_2darray(x)
+        self._snapshots = self._snapshots.squeeze()
+
+        # Check that input data is a 1D time-series
+        if np.ndim(self._snapshots) != 1:
```
fAndreuzzi, line 147: It's the same but it's more readable:
```suggestion
        if self._snapshots.ndim > 1:
```

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+        # Check that the input time-series contains enough observations
+        if n_samples < self._d:
+            msg = """The number of snapshots provided is not enough for d={}.
+Expected at least d."""
+            raise ValueError(msg.format(self._d))
+
+        # Compute hankel matrix for the input data
+        self._hankel_matrix = self._pseudo_hankel_matrix(
+            self._snapshots.reshape(1, -1)
+        )
```
fAndreuzzi, line 162: It's the same but it's more readable:
```suggestion
        self._hankel_matrix = self._pseudo_hankel_matrix(self._snapshots[None])
```

pydmd/havok.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Roughly approximate the continuous-time system matrix
+        regression_model_continuous = (
+            regression_model_discrete - np.eye(self._r)
+        ) / dt
+
+        # Save A matrix and B vector 
+        self._A = regression_model_continuous[:-1, :-1]
+        self._B = regression_model_continuous[:-1, -1].reshape(-1, 1)
```
fAndreuzzi, line 187: It's the same but it's more readable:
```suggestion
        self._B = regression_model_continuous[:-1, -1, None]
```

fAndreuzzi (commented):
Hi @sichinaga, thank you for the new DMD. I left some minor comments on the code. It looks very good, do you mind adding a few general tests, just to make sure it can run and produce reasonable results?

pydmd/havok.py
sichinaga: Unless I am being silly, `linear_embeddings` and `forcing_input` are slightly different in that `linear_embeddings` returns all of the columns of `self._embeddings` excluding the last column, while `forcing_input` returns only the last column of `self._embeddings`.

fAndreuzzi: Sorry @sichinaga my fault, I meant that my comment here is the same I wrote for `linear_embeddings`:

`This will trigger a weird error if fit() is not called`

sichinaga: Ahhh gotcha. Thank you for clarifying! :)

sichinaga:
Thank you @fAndreuzzi for all of your great feedback so far! 😃

I will begin working on some tests now. The commit should be up soon. :)

sichinaga:
Hi again @fAndreuzzi!

Just dropping in to let you know that I have now uploaded a small suite of tests for the `HAVOK` module! I've also added in a few small edits including a new attribute and some extra precautionary checks. The edits are overall quite minor though. 

With that being said, the test suite that I uploaded is arguably quite small in comparison to the other PyDMD test suites, but I hope it serves as a good starting point! Each test is documented in case you're curious about the breadth of the current batch of tests. :) I'm happy to add many more tests, especially if there are any major ones that I left out. I just implemented the ones that I felt were most relevant to the HAVOK algorithm, but please let me know if there's anything else you feel I can do to improve the quality of this PR.

Thanks again for your kind support. :)

ndem0:
Dear @sichinaga, thanks for the addition! @fAndreuzzi already did the review, so everything is fine!

I'm going to merge this PR, we can always add new tests in future, but these are enough to avoid critical issues in the development.

fAndreuzzi:
Hi @sichinaga, thanks for adding the tests, it looks like we have a very good coverage :)
Good job!

Issue #300, opened 2022-12-08
SVD Optimal Rank is 0


MSchmidt99:
Hello, I am working through a use case of DMD which is resulting in zero dynamics produced. I have traced it through to the function compute_svd in utils, where it is calculating the optimal rank of the SVD to be 0. I'm wondering which computation for omega is being applied from the paper referenced in the docstring for the function, and if there is some way to improve the precision of the computation?

To reproduce you may set X as the following:
```
X = np.array([[ 3.85371491e-04,  9.06072298e-05, -1.30414974e-04,
         1.70775663e-04,  6.51660274e-05, -9.83726696e-05,
        -2.25007236e-04, -2.74994428e-04, -2.36555641e-04,
        -1.55353240e-04, -4.44957525e-05,  2.80992388e-05,
        -5.77463248e-06, -9.26441030e-05, -1.71763714e-04,
        -2.12542023e-04, -1.58432301e-04, -3.09056611e-05,
         1.36986554e-04,  3.37938554e-04],
       [ 9.06072298e-05, -1.30414974e-04,  1.70775663e-04,
         6.51660274e-05, -9.83726696e-05, -2.25007236e-04,
        -2.74994428e-04, -2.36555641e-04, -1.55353240e-04,
        -4.44957525e-05,  2.80992388e-05, -5.77463248e-06,
        -9.26441030e-05, -1.71763714e-04, -2.12542023e-04,
        -1.58432301e-04, -3.09056611e-05,  1.36986554e-04,
         3.37938554e-04, -8.90378232e-05],
       [-1.30414974e-04,  1.70775663e-04,  6.51660274e-05,
        -9.83726696e-05, -2.25007236e-04, -2.74994428e-04,
        -2.36555641e-04, -1.55353240e-04, -4.44957525e-05,
         2.80992388e-05, -5.77463248e-06, -9.26441030e-05,
        -1.71763714e-04, -2.12542023e-04, -1.58432301e-04,
        -3.09056611e-05,  1.36986554e-04,  3.37938554e-04,
        -8.90378232e-05,  1.10772126e-04],
       [ 1.70775663e-04,  6.51660274e-05, -9.83726696e-05,
        -2.25007236e-04, -2.74994428e-04, -2.36555641e-04,
        -1.55353240e-04, -4.44957525e-05,  2.80992388e-05,
        -5.77463248e-06, -9.26441030e-05, -1.71763714e-04,
[...]

PR #301, opened 2022-12-10
FIX SVD optimal rank is 0

MSchmidt99:
Raise a warning when compute_svd under `svd_rank=0` determines an optimal rank of 0, and update the rank to 1. This can be caught and handled if needed by setting the specific warning to raise an error as follows:
```python
warnings.filterwarnings(
    'error',
    category=RuntimeWarning,
    message='SVD optimal rank is 0. The largest singular values are '
            'indistinguishable from noise. Setting rank truncation to 1.'
)
```

Addresses #300.

### DIFF
--- a/pydmd/utils.py
+++ b/pydmd/utils.py
@@ -1,5 +1,6 @@
 """Utilities module."""
 
+import warnings
 import numpy as np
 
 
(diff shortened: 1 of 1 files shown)


fAndreuzzi:
Thanks @MSchmidt99!

PR #305, opened 2023-01-01
Move plot code to `plotter.py`

fAndreuzzi:
This improves testability making at the same time test files lighter, and eases the process towards #299 
Tests on plotting are only in `test_plotter.py`, since we don't need to test separately for each DMD variant.


mtezzele:
@fAndreuzzi I'll merge after the rebase. Thanks

fAndreuzzi:
I noticed it's a breaking change in some cases, I'll work on it and let you know when it's ready @mtezzele

PR #308, opened 2023-01-12
Added BOP-DMD.

sichinaga:
Hello again to everyone on the PyDMD team!

It has been a while, but I have finally finished another potential addition to PyDMD. This module in particular has taken me quite a while, mostly because of the lengthy variable projection and bagging routines that this module required. But nonetheless, it is done and I hope that it meets the team's expectations. 😀 

In this pull request, I've implemented Optimized DMD, which essentially poses the DMD optimization as a problem that can be solved via variable projection. In doing so, the method is able to optimally handle data that is noisy and unevenly sampled in time. It should be noted however that this process requires solving a nonlinear optimization problem, which may lead to instabilities. However, it has been shown that the Optimized DMD routine can be stabilized by averaging together an ensemble of Optimized DMD models in a routine that has now been dubbed Bagging, Optimized DMD, or BOP-DMD for short. Provided below are the relevant references for anyone who may be interested.

[Optimized DMD Paper](https://arxiv.org/abs/1704.02343)
[BOP-DMD Paper](https://arxiv.org/abs/2107.10878)

The `BOPDMD` module that I have implemented provides all of the tools necessary for performing both Optimized DMD and BOP-DMD. Users can choose a variety of relevant model parameters, such as parameters that control the specifics of the variable projection routine, as well as parameters that control how the model ensemble will be built. Once fitted, the model can reconstruct the data, form predictions, and output standard DMD outputs including modes, eigenvalues, amplitudes, Atilde, and even the full system matrix A. Uncertainty quantifications for modes, eigenvalues, amplitudes, and predictions are also computed and provided in the event that BOP-DMD is performed. I have also updated `_init_.py` to include this module, in addition to providing tests and the corresponding .rst file.

Please let me know if there is anything I can do in order to improve the quality of this module. I deeply apologize for its sheer length. However, given the practicality of the method, I sincerely believe that it will be a valuable addition to the package.

Thank you all so much for all of your kind support and all of your feedback thus far. It has truly been an honor to be able to contribute to this package and work with you all, and I look forward to hopefully working on even more code together. :)

Sincerely,
Sara Ichinaga


pydmd/bopdmd.py
```diff
@@ -0,0 +1,991 @@
...
+            msg += "trial_size parameter is small enough for bagging."
+            raise ValueError(msg)
+        elif batch_size == 0:
+            msg = "Error bagging the input data. Please ensure that the "
+            msg += "trial_size parameter is large enough for bagging."
+            raise ValueError(msg)
+
+        # Obtain and return subset of the data.
+        all_inds = np.arange(H.shape[0])
+        subset_inds = np.sort(
```
fAndreuzzi, line 245: Just out of curiosity, do you actually need the `sort` here? What happens if `_bag` returns a random sample which does not take into account the relative order of the rows in `H`?

pydmd/bopdmd.py
```diff
@@ -0,0 +1,991 @@
...
+                if np.argmax(all_var) == 0:
+                    self._eig_sort = "real"
+                elif np.argmax(all_var) == 1:
+                    self._eig_sort = "imag"
+                else:
+                    self._eig_sort = "abs"
+
+            if self._eig_sort == "real":
+                sorted_inds = np.argsort(e_i)
+
```
fAndreuzzi, line 543: Please remove newlines between `if/elif/else` cases, and the other occurrences in the file as well.

pydmd/bopdmd.py
```diff
@@ -0,0 +1,991 @@
...
+            elif self._eig_sort == "imag":
+                e_i_real_imag_swapped = e_i.imag + (1j * e_i.real)
+                sorted_inds = np.argsort(e_i_real_imag_swapped)
+
+            elif self._eig_sort == "abs":
+                sorted_inds = np.argsort(np.abs(e_i))
+            else:
+                raise ValueError("Provided eig_sort method is not supported.")
+
+            all_w[i, :, :] = w_i[:, sorted_inds]
```
fAndreuzzi, line 553:
```suggestion
            all_w[i] = w_i[:, sorted_inds]
```

fAndreuzzi:
Please @sichinaga merge or rebase your branch against `master` to fix the failing tests in this PR.

pydmd/bopdmd.py
```diff
@@ -0,0 +1,991 @@
...
+        (See BOPDMD documentation for full description for each option.)
+    :type varpro_opts: list
+
+    :param use_proj: Flag that determines the type of computation to perform.
+        If True, fit input data projected onto the first svd_rank POD modes or
+        columns of proj_basis if provided. If False, fit the full input data.
+    :type use_proj: bool
+
+    :param init_alpha: Initial guess for the continuous-time DMD eigenvalues.
+    :type init_alpha: numpy.ndarray
```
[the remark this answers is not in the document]
mtezzele: @fAndreuzzi are you sure it works with `np.` instead of `numpy.`?

pydmd/bopdmd.py
```diff
@@ -0,0 +1,991 @@
...
+        rank = X.shape[1]
+
+    return rank
+
+
+class BOPDMDOperator(DMDOperator):
+    """
+    BOP-DMD operator.
+
+    :param varpro_opts: List containing all variable projection options.
```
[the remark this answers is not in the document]
sichinaga: So to give this parameter a bit more context, the desired entries of `varpro_opts` are actually passed to an initialization of a `BOPDMD` model. They are then checked as a safeguard against unreasonable variable projection parameter values, and then stored in the list `varpro_opts` for conciseness. This list is what is then passed to the `BOPDMDOperator`, so users never actually directly access, edit, or build the `varpro_opts` list.

I completely agree that a dictionary makes a lot of sense for this particular batch of parameters. However, I also implemented this in the way that I did specifically so that users could initialize the variable projection parameters upon initializing their `BOPDMD` model if they so choose. I then decided to use a list since I didn't intend for users to ever necessarily update these parameters afterwards (they would maybe only ever need to view these parameters, which is why I implemented a function that prints all of these parameters). The `varpro_opts` list is really only there for conciseness when passing the parameters from the `BOPDMD` model to the `BOPDMDOperator` object.

Alternatively, I can still implement the suggested functionality where the variable projection parameters are not passed to the `BOPDMD` model but are instead set to default values, stored in a dictionary, and updated after the model is initialized if desired. I can envision either functionality working out in the end, so I'm open to anything. :)

pydmd/bopdmd.py
sichinaga: I haven't tested this myself, so I'm not completely sure what would happen quantitatively, but without the `sort`, the original ordering of the snapshots and time points will be lost during each bagged trial. In other words, each trial will involve solving the $X_2=AX_1$ DMD problem but with the columns of $X$ shuffled rather than in their actual sequential order (assuming that each column of $X$ contains a single snapshot). Though overall, I believe it's best to preserve the snapshot order with respect to time for the sake of model consistency.

pydmd/bopdmd.py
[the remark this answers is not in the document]
fAndreuzzi: Thanks for the explanation! Now I found the code in `BOPDMD` which is doing the packing before passing the list to the operator. I think looking for conciseness for passing parameters around is very nice, but we can probably make it less error prone by passing a dictionary instead.

I suggest to require an optional `dict` called `varpro_opts` (or something similar) in `BOPDMD.__init__` with default value `None` (which should be replaced with an empty `dict` in the method body). Then in `fit` you can expand it in the call to `BOPDMDOperator.__init__` with `**self._varpro_opts`. If you decide to go this way default values should be moved to `BOPDMDOperator` constructor, such that they are used in case the expansion of `**self._varpro_opts` does not contain some values.

This way even if you change the order of the arguments in `BOPDMDOperator` the code is not broken and you don't risk bugs in case you forget to update both sides. Let me know what you think :)

pydmd/bopdmd.py
fAndreuzzi: Okay thanks @sichinaga !

pydmd/bopdmd.py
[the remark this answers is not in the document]
sichinaga: Sounds like a great idea! The functionality should now be implemented! :)

pydmd/bopdmd.py
```diff
@@ -0,0 +1,943 @@
...
+
+        # Perform num_trials many trials of optimized dmd.
+        for i in range(self._num_trials):
+            H_i, subset_inds = self._bag(H, self._trial_size)
+            w_i, e_i, b_i, _, _ = self._single_trial_compute_operator(
+                H_i, t[subset_inds], e_0
+            )
+
+            # Set the sorting style if _eig_sort is "auto".
+            if self._eig_sort == "auto":
```
fAndreuzzi, line 570: The mechanics of `self._eig_sort` are not completely clear, in the beginning `self._eig_sort` contains the user preference passed through the constructor. However if `self._eig_sort=='auto'` the value is going to be changed to one of `real, imag, abs` during the first loop of the `for` loop at L563. This way the ordering inferred in the first iteration is imposed on all the other iterations. Is this the intended behavior @sichinaga ?

pydmd/bopdmd.py
```diff
@@ -0,0 +1,943 @@
...
+                elif np.argmax(all_var) == 1:
+                    self._eig_sort = "imag"
+                else:
+                    self._eig_sort = "abs"
+
+            # Sort the results according to eigenvalue.
+            if self._eig_sort == "real":
+                sorted_inds = np.argsort(e_i)
+            elif self._eig_sort == "imag":
+                sorted_inds = np.argsort(e_i.imag + (1j * e_i.real))
```
fAndreuzzi, line 586: Is this really doing the expected sorting? Maybe you could just swap real and imaginary parts and use `np.argsort` again.

pydmd/bopdmd.py
```diff
@@ -0,0 +1,943 @@
...
+        Z = (ux2 - ux1).dot(np.linalg.inv(T))
+        U, s, V = compute_svd(Y, self._svd_rank)
+        S = np.diag(s)
+
+        # Compute the matrix Atilde and return its eigenvalues.
+        Atilde = np.linalg.multi_dot([U.conj().T, Z, V, np.linalg.inv(S)])
+
+        return np.linalg.eig(Atilde)[0]
+
+
```
fAndreuzzi, line 732:
```suggestion
```

fAndreuzzi: (additional newline)

pydmd/bopdmd.py
sichinaga: I realize that this implementation may not be the most optimal/clear, but in short, the answer to your question is **yes.**

Basically, the goal of `eig_sort` is to ensure that as we create multiple DMD models and average out their eigenvalues, modes, etc., we don't accidentally average together the wrong attributes. For example, if the ground truth eigenvalues are $i, -i$, we tend to see computed DMD eigenvalues in the neighborhood of $x+i, y-i$ for some extremely small $x,y$ values. So as we average out these eigenvalues across DMD models, we most likely want to consistently sort the computed eignvalues according to imaginary component before averaging.

I thought about removing `real, imag, abs` as options altogether to simply make `auto` the only available functionality, but I figured allowing users to choose one of the three sorting methods might be nice, especially in the event that the user is aware of the expected eigenvalue spectra and/or there is no best sorting option.

Let me know if you have any thoughts though! I'd be happy to make improvements, especially since this portion of the code was somewhat improvised. :)

pydmd/bopdmd.py
sichinaga: I believe it should work as expected? Though for the sake of readability, I have changed it so that the real and imaginary parts are swapped before `np.argsort` is called.

pydmd/bopdmd.py
fAndreuzzi: Okay I thing I got it, thanks for the explanation @sichinaga :)

pydmd/bopdmd.py
fAndreuzzi: Yeah indeed I misread 👍

fAndreuzzi:
Hi @sichinaga, could you please apply this patch to your branch with `git apply codacy-fixes.patch` ? 
After commit and push I think we're ready to merge 👍 

[codacy-fixes.patch](https://github.com/mathLab/PyDMD/files/10514158/codacy-fixes.patch)

PR #319, opened 2023-01-16
Export tutorials in CI

fAndreuzzi:
Export automatically to `.py` and `.html` tutorials changed in any commit to `master`.

### DIFF
--- a/.github/workflows/export_tutorials.yml
+++ b/.github/workflows/export_tutorials.yml
@@ -1,0 +1,85 @@
+name: "Export tutorials"
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'tutorials/**/*.ipynb'
+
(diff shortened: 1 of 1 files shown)


fAndreuzzi:
Example commit (from my fork): https://github.com/fAndreuzzi/PyDMD/commit/b95e349410ff5c18ce9ec3b1602bc1bf457beac8

fAndreuzzi:
@ndem0 even though the HTML file is changed, it looks like it's not updaed in mathlab.github.io, do I need to run Sphinx?

mtezzele:
> @ndem0 even though the HTML file is changed, it looks like it's not updaed in mathlab.github.io, do I need to run Sphinx?

I think the online documentation is updated on a monthly basis when a new tag is pushed or with the corresponding action.

PR #325, opened 2023-01-28
Added randomized matrix parameter to rdmd

sichinaga:
In this PR, I've added `rand_mat` as a parameter to `RDMD` so that users may pass their `RDMD` model a random matrix of their choice for the Randomized QB Decomposition.

Basically, the Randomized QB Decomposition begins with the computation of $X \Omega$, with $X$ being the data matrix and $\Omega$ being a random matrix that is typically drawn from a normal Gaussian. The issue is that computing $\Omega$ can be expensive, so it would make sense that users might want to pre-compute $\Omega$ and pass it to the `RDMD` model, especially if the user aims to train multiple `RDMD` models.

Another benefit of this additional parameter is that it gives users the freedom to use a random matrix other than one drawn from a Gaussian if they so desire. The authors of the RDMD paper somewhat hint at the existence of random matrices that are better than ones drawn from a Gaussian, which could potentially be implemented in `RDMD` in the future, but for now, I think that the freedom to provide the desired random matrix for the Randomized QB Decomposition is sufficient.

Let me know if any questions, comments, or concerns arise with this alteration to `RDMD` :)

### DIFF
--- a/pydmd/rdmd.py
+++ b/pydmd/rdmd.py
@@ -54,14 +54,22 @@ class RDMD(CDMD):
     """
     Randomized Dynamic Mode Decomposition
 
-    :param int oversampling: Number of additional samples to use when
-        computing the random test matrix.
-        Note that oversampling = {5,10} is often sufficient.
-    :param int power_iters: Number of power iterations to perform.
-        Note that power_iters = {1,2} leads to considerable improvements.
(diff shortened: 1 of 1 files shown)


fAndreuzzi:
Ready to merge for me

PR #326, opened 2023-02-01
Added diagnostics plotter

sichinaga:
Hi!

In this PR, I've implemented a plotting function called `diagnostics`, which, given a fitted DMD model, will generate a 3x3 summarizing plot that contains the following components:

Plot 1: The singular value spectrum for the data used to fit the model.
Plots 2-3: The discrete-time and continuous-time DMD eigenvalues.
Plots 4-6: DMD modes of the user's choice (which can be 1D or 2D).
Plots 7-9: Dynamics corresponding with the plotted modes.

A sample figure is provided below, which was built via the following (here, my data $X$ contains flow past a cylinder data). 
```
dmd = DMD(svd_rank=15)
dmd.fit(X)
diagnostics(dmd, snapshots_shape=(nx,ny), index_modes=[0,1,3], order="F")
```
Note that the singular values and eigenvalues are color-coordinated as to highlight associations with the plotted modes and dynamics. Also note that the sizes of the eigenvalue markers are scaled according to the amount of data variance captured by that eigenvalue/mode pairing, i.e. if eigenvalue marker B is half the size of eigenvalue marker A, this indicates that the eig/mode pairing given by A captures two times as much data variance as eig/mode pairing B.

I've also done my best to reuse parameter values that are already incorporated into other `plotter.py` functions for the sake of consistency, such as `snapshots_shape`, `filename`, `order`, and `figsize` to name a few. But of course, there are several parameters that I have omitted mostly for the sake of simplicity, such as marker sizes, linewidths, whether to include the unit circle or not, the number of singular values to plot (which is currently set to be at most 50), etc. Whether to increase the amount of parameters for the method is still up for debate to me, but overall, I've chosen parameters that I believe make practical sense.

Overall, this functionality is meant to be a convenience function that will allow users to very easily and efficiently visualize the results of whatever DMD model they may have built with PyDMD. I think that it will be a very valuable addition to the plotting functionality of PyDMD and I'd be more than happy to consider any questions or feedback that you might have. 😃 

(I also think my most recent RDMD commit snuck its way into this PR for some reason, so I apologize for that😅)

cylinder

### DIFF
--- a/pydmd/plotter.py
+++ b/pydmd/plotter.py
@@ -528,7 +528,7 @@ def plot_snapshots_2D(
         plt.show()
 
 
-def diagnostics(
+def plot_summary(
     dmd,
     snapshots_shape=None,
     index_modes=None,
@@ -626,7 +626,7 @@ def diagnostics(
     # Only plot (at most) 50 of the leading singular values.
     s_var = s_var[:50]
 
-    # Generate the diagnostics plot.
+    # Generate the summarizing plot.
     fig, (eig_axes,
           mode_axes,
           dynamics_axes) = plt.subplots(3, 3, figsize=figsize, dpi=200)
--- a/pydmd/plotter.py
+++ b/pydmd/plotter.py
@@ -535,7 +535,7 @@ def diagnostics(
     filename=None,
     order="C",
     figsize=(12,8),
-    mode_colors=["r","b","g","gray"],
+    mode_colors=None,
 ):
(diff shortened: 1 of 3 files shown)


mtezzele:
@sichinaga This is great! I love the plots. I would suggest changing the name to `plot_summary` instead of `diagnostics`. This is also a great material for one of the first tutorials.

PR #327, opened 2023-02-06
Added compute_A parameter to BOPDMD

sichinaga:
Hi! This PR just involves a tiny (yet crucial) edit to the `BOPDMD` module, which is the addition of a `compute_A` flag.

Previously, I programmed the `BOPDMD` module so that it computes the full Koopman operator $A$ all of the time. However, this can get problematic, especially in the event where one's data is sufficiently high-dimensional and $A$ becomes prohibitively expensive to compute. Thus, I've added the `compute_A` parameter to the module, which allows users to choose whether or not they would like the module to compute $A$, since computing $A$ still has potential to be helpful, especially when one is dealing with data from small systems of ODEs or just smaller datasets in general.

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -28,6 +28,9 @@ class BOPDMDOperator(DMDOperator):
     """
     BOP-DMD operator.
 
+    :param compute_A: Flag that determines whether or not to compute the full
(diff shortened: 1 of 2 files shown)


mtezzele:
@sichinaga could you please add a test where the flag is false?

PR #329, opened 2023-02-09
Added PiDMD

sichinaga:
Hello PyDMD team! 😃 

Here is yet another potential addition to the PyDMD family of modules: [Physics-informed DMD](https://arxiv.org/abs/2112.04307).

In short, Physics-informed DMD (PiDMD) simply replaces the DMD regression scheme

$A = \text{argmin}_{\text{rank}(A)=r} ||Y-AX||_F$

with 

$A = \text{argmin}_{A \in \mathcal{M}} ||Y-AX||_F$

where $\mathcal{M}$ denotes some user-chosen matrix manifold that contains a family of matrices, such as unitary matrices, symmetric matrices, etc. In doing so, one has the option to constrain the structure of the DMD matrix $A$ in the event that the user possesses some knowledge or intuition regarding the physics of their dataset and therefore the structure of the operator $A$. In doing so, one has potential to create DMD models that are more physically accurate as well as robust to measurement noise.

In my implementation of the `PiDMD` module, the module is extended from the `DMD` class since the only difference between the two modules is the execution of the `compute_operator` function. For `PiDMD` in particular, `compute_operator` is overwritten so that it involves solving the specific Procrustes problem that is associated with the inputted matrix manifold. Currently, 19 matrix manifolds may be evaluated by the module, with each manifold having a corresponding test within the `test_pidmd.py` file.

Additionally as a minor edit, I've edited the `compute_rank` function so that in the event that `svd_rank=-1` the rank that is returned is $\text{min}(m,n)$ for a data matrix $X \in \mathbb{R}^{m \times n}$, mostly because this seems to be more reflective of the chosen rank in the event that no truncation is desired.

Thank you so much for your kind and generous support in this endeavor to build upon this already amazing package. And as always, feel free to follow up with me on anything regarding this PR and please let me know if there's anything that I can do to get this PR up to standards. :)

Sincerely,
Sara Ichinaga

### DIFF
--- a/pydmd/pidmd.py
+++ b/pydmd/pidmd.py
@@ -124,7 +124,7 @@ class PiDMDOperator(DMDOperator):
         if manifold_opt is None:
             ind_mat = np.ones((nx, 2), dtype=int)
         elif isinstance(manifold_opt, int) and manifold_opt > 0:
-            ind_mat = manifold_opt * np.ones((nx, 2))
+            ind_mat = manifold_opt * np.ones((nx, 2), dtype=int)
         elif (isinstance(manifold_opt, tuple)
                 and len(manifold_opt) == 2
                 and np.all(np.array(manifold_opt) > 0)):
--- a/pydmd/pidmd.py
+++ b/pydmd/pidmd.py
@@ -89,36 +89,34 @@ class PiDMDOperator(DMDOperator):
 
     def _compute_unitary(self, X, Y):
         """
-        Given the data matrices X and Y, solves the for the best-fit unitary
-        operator A that solves the relationship Y=AX. Returns the corresponding
+        Given the data matrices X and Y, solves for the best-fit unitary
+        operator A that solves the relationship Y=AX. Stores the corresponding
         reduced operator atilde.
         """
         Ux = compute_svd(X, self._svd_rank)[0]
         Yproj = Ux.conj().T.dot(Y)
         Xproj = Ux.conj().T.dot(X)
         Uyx, _, Vyx = compute_svd(Yproj.dot(Xproj.conj().T), -1)
-        atilde = Uyx.dot(Vyx.conj().T)
-        return atilde
+        self._Atilde = Uyx.dot(Vyx.conj().T)
 
 
     def _compute_uppertriangular(self, X, Y):
         """
-        Given the data matrices X and Y, solves for and returns the best-fit
+        Given the data matrices X and Y, solves for and stores the best-fit
         uppertriangular matrix A that solves the relationship Y=AX.
         """
         self._check_compute_A()
         R, Q = rq(X, mode="economic")
         Ut = np.triu(Y.dot(Q.conj().T))
-        A = np.linalg.lstsq(R.T, Ut.T, rcond=None)[0].T
-        return A
+        self._A = np.linalg.lstsq(R.T, Ut.T, rcond=None)[0].T
 
 
     def _compute_diagonal(self, X, Y, manifold_opt):
         """
         Given the data matrices X and Y and specification for the diagonals of
-        the matrix A, solves for and returns the best-fit matrix A that solves
+        the matrix A, solves for and stores the best-fit matrix A that solves
         the relationship Y=AX and has diagonal entries specified by
-        manifold_opt. Only the eigenvalues and eigenvectors of A are returned
+        manifold_opt. Only the eigenvalues and eigenvectors of A are stored
         if the compute_A flag is not True.
         """
         # Specify the index matrix for the diagonals of A.
@@ -159,18 +157,19 @@ class PiDMDOperator(DMDOperator):
         A_sparse = sparse.coo_array(
             (np.hstack(R), (np.hstack(I), np.hstack(J))), shape=(nx,nx)
         )
-
         if self._compute_A:
-            return None, None, A_sparse.toarray()
-
-        return sparse.linalg.eigs(A_sparse, k=nx-2), None
+            self._A = A_sparse.toarray()
+        else:
+            self._eigenvalues, self._modes = sparse.linalg.eigs(
+                A_sparse, k=compute_rank(X, self._svd_rank)
+            )
 
 
     def _compute_symmetric(self, X, Y, skewsymmetric=False):
         """
-        Given the data matrices X and Y, solves the for the best-fit symmetric
+        Given the data matrices X and Y, solves for the best-fit symmetric
         (or skewsymmetric) operator A that solves the relationship Y=AX.
-        Returns the corresponding reduced operator atilde.
+        Stores the corresponding reduced operator atilde.
         """
         U, s, V = compute_svd(X, -1)
         C = np.linalg.multi_dot([U.conj().T, Y, V])
@@ -192,16 +191,17 @@ class PiDMDOperator(DMDOperator):
                     atilde[i, j] /= (s[i] ** 2 + s[j] ** 2)
             atilde += atilde.conj().T - np.diag(np.diag(atilde.real))
 
-        return atilde
+        self._Atilde = atilde
 
 
     def _compute_toeplitz(self, X, Y, flipped=False):
         """
-        Given the data matrices X and Y, solves for and returns the
+        Given the data matrices X and Y, solves for and stores the
         best-fit toeplitz (or hankel if flipped=True) operator A that
         solves the relationship Y=AX.
         """
         self._check_compute_A()
+
         nx, nt = X.shape
         if flipped: # hankel
             J = np.fliplr(np.eye(nx))
@@ -230,17 +230,15 @@ class PiDMDOperator(DMDOperator):
         new_A = ifft(fft(np.diag(d), axis=0).T, axis=0).T
 
         # Extract toeplitz matrix from the circulant matrix.
-        A = new_A[:nx, :nx].dot(J)
-
-        return A
+        self._A = new_A[:nx, :nx].dot(J)
 
 
     def _compute_circulant(self, X, Y, circulant_opt):
         """
-        Given the data matrices X and Y, solves for and returns the best-fit
+        Given the data matrices X and Y, solves for and stores the best-fit
         circulant matrix A that solves the relationship Y=AX and satisfies any
         additional conditions set by circulant_opt. Only the eigenvalues and
-        eigenvectors of A are returned if the compute_A flag is not True.
+        eigenvectors of A are stored if the compute_A flag is not True.
         """
         nx = len(X)
         fX = fft(X, axis=0)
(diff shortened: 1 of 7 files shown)


sichinaga:
There also appears to be some testing issues because the `PiDMD` module uses `scipy.sparse.coo_array`. Is this something that I should attempt to edit out? Or will this issue be fixed with an update to the `scipy` package?

pydmd/pidmd.py
```diff
@@ -0,0 +1,535 @@
...
+                        )[-1]
+                    d.append(sol)
+
+                bd = block_diag(*d)
+                fI = fft_block(np.eye(nx))
+                A = fft_block(bd.dot(fI).conj()).conj()
+        else:
+            raise ValueError("The selected manifold is not implemented.")
+
+        return eigenvalues, modes, atilde, A
```
ndem0, line 397: This function is extremely complex to read and understand. Instead of having this huge `if ... elif ...` list, can you break it into smaller private functions (one for any method)?

pydmd/pidmd.py
```diff
@@ -0,0 +1,535 @@
...
+                x = np.linalg.lstsq(C, b, rcond=None)[0].T
+                I.append(j * np.ones(l2 - l1))
+                J.append(np.arange(l1, l2))
+                R.append(x)
+
+            # Build A as a sparse matrix.
+            I_mat = np.hstack(I)
+            J_mat = np.hstack(J)
+            R_mat = np.hstack(R)
+            A_sparse = sparse.coo_array((R_mat, (I_mat,J_mat)), shape=(nx,nx))
```
ndem0, line 172: I think `coo_array` has been added in scipy version 1.8, maybe try to add the version requirement (`>=1.8`) at this line https://github.com/mathLab/PyDMD/blob/63e5c326eb8b4878faa650656b5b32ffc3112a74/setup.py#L17

mtezzele:
Hi @sichinaga, thank you for the addition. I suggest adding in the documentation all the possible manifolds the user can pass as argument.

pydmd/pidmd.py
sichinaga: Hi Nicola! My apologies for keeping the code in such a chaotic state. I just added in separate functions for each distinct Procrustes problem. The `if ... elif ...` structure is still in place, but different private functions are executed depending on the chosen manifold. Is this the change that you had in mind?

pydmd/pidmd.py
sichinaga: Thanks for the tip Nicola! I implemented this change, but it looks like we traded the original error for `ERROR: No matching distribution found for scipy>=1.8`.

ndem0:
Hi @sichinaga, it's a great job!
Personally I think the list of strings you put in docstring is enough, if we also want to put a small description for all the cases the entire page becomes too long.
There are some small "coding trick" to avoid potential bugs in future, I gonna continue to comment directly in the code! But we are close to merging!

pydmd/pidmd.py
```diff
@@ -0,0 +1,624 @@
...
+        Computes and returns either (1) the eigenvalues and eigenvectors of A,
+        (2) the reduced operator atilde, or (3) the full operator A, depending
+        on the chosen manifold and the compute_A parameter.
+
+        :return: eigenvalues of A, eigenvectors of A, reduced operator atilde,
+            and the full operator A. None is returned for quantities that are
+            not computed.
+        :rtype: [numpy.ndarray, numpy.ndarray, NoneType, NoneType],
+            [NoneType, NoneType, numpy.ndarray, NoneType], or
+            [NoneType, NoneType, NoneType, numpy.ndarray]
```
ndem0, line 424: Even if the method is private, I would try to make the returning type easier to mantain in the future. Instead of having this list of 4 different 'objects' that can be None, what if we put only the real computed matrices on a dictionary. Example:
`return {'atilde': atilde}  # if only atilde is computed`
`return {'modes': modes, 'eigenvalues': eigenvalues}  # when modes and lambda`
and so on.
In this way, in `compute operator` we have just to consider the three cases in a more readable way.

sichinaga: Hi @ndem0. Thank you so much for the feedback. 😃

So, I didn't implement this exact structure, but I did make some changes that I think help a lot with this readability issue.

So instead of having all of the helper functions return values like `atilde` etc., I made it so that the helper functions simply update the `PiDMDOperator` attributes like `self._Atilde`, just because there's so much variability across the Procrustes problem solutions when it comes to which DMD attributes are computed.

Let me know if you think this can be improved upon. :)

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+
+        # Keep track of info for building A as a sparse coordinate matrix.
+        I = []
+        J = []
+        R = []
+
+        # Solve min||Cx-b|| along each row.
+        for j in range(nx):
+            l1 = int(max(j-(ind_mat[j,0]-1), 0))
+            l2 = int(min(j+(ind_mat[j,1]-1), nx-1) + 1)
```
fAndreuzzi, line 149:
```suggestion
        nxs = np.arange(nx)
        l1s = (nxs - ind_mat[:, 0] + 1).clip(min=0)
        l2s = (nxs + ind_mat[:, 1]).clip(max=nx)
        for j in range(nx):
                l1, l2 = l1s[j], l2s[j]
```

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+        Given the data matrices X and Y and specification for the diagonals of
+        the matrix A, solves for and stores the best-fit matrix A that solves
+        the relationship Y=AX and has diagonal entries specified by
+        manifold_opt. Only the eigenvalues and eigenvectors of A are stored
+        if the compute_A flag is not True.
+        """
+        # Specify the index matrix for the diagonals of A.
+        nx = len(X)
+        if manifold_opt is None:
+            ind_mat = np.ones((nx, 2))
```
fAndreuzzi, line 125:
```suggestion
            ind_mat = np.ones((nx, 2), dtype=int)
```

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+        elif isinstance(manifold_opt, int) and manifold_opt > 0:
+            ind_mat = manifold_opt * np.ones((nx, 2))
```
fAndreuzzi, line 127:
```suggestion
            ind_mat = manifold_opt * np.ones((nx, 2), dtype=int)
```

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+    def _compute_symmetric(self, X, Y, skewsymmetric=False):
+        """
+        Given the data matrices X and Y, solves for the best-fit symmetric
+        (or skewsymmetric) operator A that solves the relationship Y=AX.
+        Stores the corresponding reduced operator atilde.
+        """
+        U, s, V = compute_svd(X, -1)
+        C = np.linalg.multi_dot([U.conj().T, Y, V])
+        r = compute_rank(X, self._svd_rank)
+        atilde = np.zeros((r, r), dtype="complex")
```
fAndreuzzi, line 177:
```suggestion
```

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+        if skewsymmetric:
+            for i in range(r):
+                atilde[i, i] = 1j * (C[i, i].imag / s[i])
+                for j in range(i + 1, r):
+                    atilde[i, j] = -s[i] * np.conj(C[j,i]) + s[j] * C[i,j]
+                    atilde[i, j] /= (s[i] ** 2 + s[j] ** 2)
+            atilde += -atilde.conj().T - 1j * np.diag(np.diag(atilde.imag))
+        else: # symmetric
+            for i in range(r):
+                atilde[i, i] = C[i, i].real / s[i]
```
fAndreuzzi, line 188:
```suggestion
            atilde = np.diag(np.diagonal(C).real / s)
            for i in range(r):
```

pydmd/pidmd.py
```diff
@@ -0,0 +1,609 @@
...
+        Stores the corresponding reduced operator atilde.
+        """
+        U, s, V = compute_svd(X, -1)
+        C = np.linalg.multi_dot([U.conj().T, Y, V])
+        r = compute_rank(X, self._svd_rank)
+        atilde = np.zeros((r, r), dtype="complex")
+
+        if skewsymmetric:
+            for i in range(r):
+                atilde[i, i] = 1j * (C[i, i].imag / s[i])
```
fAndreuzzi, line 181:
```suggestion
            atilde = 1j * np.diag(np.diagonal(C).imag / s)
            for i in range(r):
```

fAndreuzzi (commented):
Hi @sichinaga, a few comments from me. Thanks 😄

fAndreuzzi:
Also, I propose to make `pidmd.py` lighter by moving all the functions like `_compute_symmetric`, `_compute_diagonal`, `_compute_toeplitz` that are not strictly related to DMD into a separate file (I propose `pidmd_utils.py`). I think each function could return a dict `{"matrix": ..., "modes": ...., "eigs": ...}`, possibly with some entries set to `None` depending on what you compute. Then you could extract from this dict the fields needed in `PiDMDOperator`. 

I like this more than delegating to individual methods which field to set (hard to test and error-prone), and this would also make the file much cleaner.

What do you think @ndem0 @mtezzele @sichinaga ?

sichinaga:
Hi @fAndreuzzi!

Thanks for all of the great feedback. 😄 

I have now incorporated all of your suggestions into the code, including the suggestion to build a `pidmd_utils.py` file. The latest commit should contain all of the suggested changes. Let me know if there are any additional issues with the code!

pydmd/pidmd_utils.py
```diff
@@ -0,0 +1,327 @@
...
+    for j in range(nx):
+        l1, l2 = l1s[j], l2s[j]
+        C = X[l1:l2].T
+        b = Y[j].T
+        I.append(j * np.ones(l2 - l1))
+        J.append(np.arange(l1, l2))
+        R.append(np.linalg.lstsq(C, b, rcond=None)[0].T)
+
+    # Build A as a sparse matrix.
+    A_sparse = sparse.coo_array(
```
fAndreuzzi, line 86: `coo_matrix` should be practically equivalent if you just use `A_sparse` to compute the eigenvalues:
```suggestion
    A_sparse = sparse.coo_matrix(
```

pydmd/pidmd.py
fAndreuzzi: @sichinaga I propose you remove the `>=1.8` and apply my latest suggestion about `coo_matrix`, which existed before `coo_array`.

fAndreuzzi:
Looks very good @sichinaga! For me we can merge after we sort out the SciPy version problem.

pydmd/pidmd_utils.py
mtezzele: If I am not mistaken `sparse.coo_matrix` does not have the same constraints as `sparse.coo_array` for the `scipy` version. In that case we can remove such a constraint, otherwise we will abandon py3.7, which will be done anyway, eventually.

fAndreuzzi:
I think this is ready to merge @ndem0 @mtezzele @sichinaga

jacktang:
Hello @sichinaga,

Could you please provide some examples about piDMD?

PR #336, opened 2023-02-23
BOP-DMD tutorial

klapo:
I hope it is ok that I wrote a tutorial for the BOP-DMD. I just lifted the example from Tutorial 2 since that gives a simple noisy toy data set that provides a direct comparison between the optDMD/BOPDMD and the exact DMD.

I also added the `eigenvalue_std` and `amplitude_std` to the `BOPDMD` class, more easily exposing the measure of uncertainty to the user.

Looking forward to feedback on how to improve the tutorial or better integrate it!

Changes relative to Tutorial 2:
- Updated the numpy random number generator to use the recommended interface. Specified the random seed for reproducibility
- Made everything numpy arrays, separately specifying the 2d snapshots for plotting and the 1d snapshots for fitting
- Specficied the noise amplitude as 0.5 instead of 0.4
- Moved to using the spatial mean instead of the two dimensional integral for showing the time dynamics and forecasting.

One last question. There are now two "optimized DMDs" in PyDMD. Is there a preferred nomenclature for avoiding ambiguities on which optimized version is being used?


mtezzele:
@klapo Many thanks for the contribution! 
@sichinaga could you please help us in reviewing this PR?

mtezzele:
I think the tutorial is great. I kindly ask @klapo to add the tutorial to the readme table (in the `tutorials` folder) and generate the `py` and `html` version. You should be able to export the notebook in different file formats. 
If I may I suggest adding at the end of the tutorial an open question for students approaching the tutorial. For example let them try different comparisons with other DMD variants created to deal with noisy data...

mtezzele:
@klapo also please do a rebase and rerun of the tutorial, just to avoid conflicts.

sichinaga:
Hi @klapo! Thank you for the PR! The tutorial looks awesome so far. 😄 My apologies for the delayed review, but I will begin providing feedback shortly. The additional properties in the `BOPDMD` class look great by the way!

I will be providing Jupyter Notebook feedback by commenting on the `tutorial-14-bop-dmd.ipynb` file. Apologies in advance for not being able to comment directly in the notebook, but I will do my best to cite the appropriate code cells.

Additionally, to address your question about differentiating between the two optDMDs, I think it's worth noting that the optDMD that is implemented in the `OptDMD` module (which refers to the work found [here](https://arxiv.org/abs/1610.02962)) computes the optimal closed-form solution to the $Y=AX$ exact DMD fitting problem. Meanwhile, the optDMD used by BOP-DMD computes the optimal solution to the exponential fitting problem $X=\Phi \text{diag}(b)T(\omega)$ and is optimal is the sense that it suppresses noise bias by solving for the DMD diagnositics directly. 

Hence, one distinction that I would suggest is to explicitly refer to the first optDMD as the "optimal closed-form exact DMD solution" or possibly "optimal exact DMD" for short, while referring to the second optDMD as "optimized DMD," as these names would be consistent with the literature.

tutorials/tutorial14/tutorial-14-bop-dmd.ipynb
sichinaga, line 1: **Cell 1** contains some unused imports. Consider using the following code.
```
import copy
import numpy as np
import matplotlib.pyplot as plt

from pydmd.bopdmd import BOPDMD
from pydmd import DMD
```

sichinaga, line 1: Consider using more descriptive variable names in the `relative_error` function in **cell 2**.
```
def relative_error(x_est, x_true):
    return np.linalg.norm(x_est - x_true) / np.linalg.norm(x_true)
```

sichinaga, line 1: The method descriptions under the **Fitting and forecasting** header are a bit confusing. In reference to my [previous comment](https://github.com/mathLab/PyDMD/pull/336#issuecomment-1471011697) on distinguishing between the two optDMD methods, I would suggest the following clarifications:

1. exact DMD with the **optimal closed-form solution**
2. **optimized DMD (note, different from optimal exact DMD)**
3. BOP-DMD

sichinaga, line 1: In reference to the code following the **Fitting and forecasting** header, I do not believe that
```
dmd = DMD(svd_rank=1, tlsq_rank=2, exact=True, opt=True)
```
produces the model that you describe in (1) in the preceding markdown cell. I believe ticking `opt=True` in a `DMD` model actually influences how the DMD amplitudes are computed and it does _not_ influence whether or not the optimal closed-form solution is computed [[opt doc](https://mathlab.github.io/PyDMD/dmdbase.html)]. I believe the following code produces the model that you are describing in (1) [[OptDMD doc](https://mathlab.github.io/PyDMD/optdmd.html)]. Of course, the model that you have currently is still a valid point of comparison. It just isn't exact DMD with the optDMD solution.
```
dmd = OptDMD(svd_rank=1, tlsq_rank=2, opt=True)
```
@fAndreuzzi @ndem0 @mtezzele please correct me if I'm mistaken.

sichinaga, line 1: Please call `plt.show()` in each plotting cell. It's quite minor, but it cleans the plot cells up a little bit. :)

sichinaga, line 1: In the first plot under the **Time Dynamics** header, it's quite difficult to see where the bopdmd line is. Although it's position is made explicit in the markdown cells, consider possibly using the `'--'` linestyle for optdmd and the `'-.'` linestyle for the true dynamics for extra visibility and clarity.
```
plt.plot(time, optdmd.dynamics.real.T / optdmd.amplitudes.real, '--', label='optdmd')
plt.plot(time, np.exp(true_eigenvalues * time).real, 'k-.', label='True dynamics')
```

klapo:
Thank @sichinaga and @mtezzele for the feedback! I'll get on it this week.

tutorials/tutorial14/tutorial-14-bop-dmd.ipynb
sichinaga, line 1: In your [initial comment](https://github.com/mathLab/PyDMD/pull/336#issue-1596965952), you stated that you specficied the noise amplitude to be 0.5, but I think you actually define it to be 1.0 with `sigma = 1` in the first code block underneath the **Setting up the toy data** header unless I'm mistaken. I'm not sure which noise amplitude you intended on using, **but I think that using a noise amplitude of 0.5 would actually be better for this tutorial.** 

I'm mostly bringing this up because in your final batch of cells, you obtain relative errors in the neighborhood of >0.5, which in general is quite high for relative error. I overall think this is due to `sigma = 1` being quite a high noise magnitude, especially considering how `data_clean.max()` is `1.9971186173351265+0j` and `data_forecast.max()` is `0.47075044308304503-1.4418403291149909e-15j` (i.e. a noise magnitude of 1 is greater than or on the same order of magnitude as the maximum value of the actual data itself, which is a very large amount of noise pollution).

sichinaga: As an additional comment on this, I found that if one uses enough BOP-DMD trials (say 100 trials) for this particular dataset, the BOP-DMD forecast error becomes ~0.29, while the DMD forecast error becomes ~0.31 with `sigma = 0.5`.

sichinaga:
Thanks again @klapo for the great PR! 😄 I just left a couple of last-minute comments on the tutorial and I am finished reviewing for the time being. Happy to review more as the edits progress. :)

tutorials/tutorial14/tutorial-14-bop-dmd.ipynb
klapo: Well this is quite confusing! If I understood everything there are _three_ `optimized` designations floating around in PyDMD meaning different things?
1) The exact DMD with the optimal closed-form solution from Heas and Herzet (2016)
2) The base DMD class with the amplitudes computed as in (1)
3) The optimized DMD from Askham and Kutz (2018)

klapo: Thanks for bringing this up! I deal with _very_ noisy data and had set sigma to be large out of curiosity and forgot to change it back.

I set the noise amplitude to be 0.4 (instead of the suggested 0.5) to match Tutorial 2 exactly. In this case, we get the following relative errors (RE):

| Method  | Forecast RE  | Spatial Modes RE  |
|---|---|---|
| Exact DMD  | 0.25  | 0.38  |
| BOP-DMD (100 trials)  | 0.24  | 0.17  |

What do you think? My preference is to use `sigma = 0.4` for consistency.

klapo: Used the suggested more descriptive variable names.

klapo: I _think_ I've now clarified the different methods correctly in the introduction of the tutorial (moved to the introduction from the cell where the methods are fit).

klapo: Fixed!

klapo: I believe that the plots now reflect your feedback and each point is distinguishable.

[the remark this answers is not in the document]
klapo: I added some descriptive comments throughout and not just in this section.

klapo:
Thank you again @sichinaga for all the great feedback!

mtezzele:
Thank you @sichinaga and @klapo for the great work. I think we can merge this PR. Could you confirm that, please?

mtezzele:
@klapo could you please rebase this branch?

tutorials/tutorial14/tutorial-14-bop-dmd.ipynb
sichinaga: Somewhat? I believe that for (2), the amplitudes are actually computed as in Sparsity-promoting DMD, as mentioned in the documentation of the `_compute_amplitudes` function, which I've provided here for reference (starting at line 695 of [dmdbase.py](https://github.com/mathLab/PyDMD/blob/edd03b5d233f9d7e607040023b4fa93e314df5de/pydmd/dmdbase.py)). Basically for (2), I believe "opt" is used to refer to the computation of optimal amplitudes rather than the computation of an optimal DMD operator (which is confusing, I know 😅). Basically to avoid any additional confusion, I would just avoid mentioning the optimal amplitudes altogether.
```
def _compute_amplitudes(self):
    """
    Compute the amplitude coefficients. If `self._opt` is False the
    amplitudes are computed by minimizing the error between the modes and
    the first snapshot; if `self._opt` is True the amplitudes are computed
    by minimizing the error between the modes and all the snapshots, at the
    expense of bigger computational cost.

    This method uses the class variables self.snapshots (for the
    snapshots), self.modes and self.eigs.

    :return: the amplitudes array
    :rtype: numpy.ndarray

    References for optimal amplitudes:
    Jovanovic et al. 2014, Sparsity-promoting dynamic mode decomposition,
    https://hal-polytechnique.archives-ouvertes.fr/hal-00995141/document
    """
```

sichinaga: Looks good to me! :)

sichinaga: This looks great! Thanks for the update :)

sichinaga: So, I noticed that the DMD model you're comparing BOP-DMD and Optimized DMD to is currently
```
dmd = DMD(svd_rank=1, tlsq_rank=2, exact=True, opt=True)
```
which as I mention in [this comment](https://github.com/mathLab/PyDMD/pull/336#discussion_r1151031279) is not exactly related to Heas and Herzet's optimal solution unless I'm mistaken. I think it would still be valuable to point out that when we say "optimized DMD", we aren't referring to Heas and Herzet, so maybe this slight alteration would suffice? Assuming you wish to keep the above model.

1. The **total least-squares DMD with optimal amplitudes**
2. The optimized DMD from Askham and Kutz (2018; **note the namespace conflict with Heas and Herzet's (2016) optimal closed-form solution!**)
3. The BOP-DMD

sichinaga:
@mtezzele @klapo I've left a couple more comments about some minor wording changes in the markdown, but other than that, I think we're ready to merge! :)

tutorials/tutorial14/tutorial-14-bop-dmd.ipynb
klapo: Thank you again! I've adapted the wording exactly as suggested.

klapo:
I rebased and changed the kernel to be `Python 3 (ipykernel)` like for the other tutorials to hopefully fix the failing errors. Is there a way to run the github actions CI locally to check for any other errors like these?

mtezzele:
> I rebased and changed the kernel to be `Python 3 (ipykernel)` like for the other tutorials to hopefully fix the failing errors. Is there a way to run the github actions CI locally to check for any other errors like these?

Unfortunately I think the only way is to execute the same list of commands, but with all the possible combinations it is just impossible...

Issue #367, opened 2023-03-31
Unintended behavior of type checking in svd_compute and compute_rank


klapo:
**Describe the bug**
For the BOPDMD module (possibly others) when setting `svd_rank=0` the resulting `proj_basis` is the wrong size. This is caused by the line `rank = np.sum(s > tau)` (in both `rdmd.compute_rank` and `utils.compute_svd`) which returns an `numpy.int64` type instead of an `int` data type. As a result in the `bopdmd.fit()` function, the`utils.compute_svd` check for `svd_rank >= 1 and isinstance(svd_rank, int)` should be `True` but is `False`. As a result, `compute_svd` falls back to the else statement and returns the length of the time dimension and overriding the optimally computed `svd_rank`.

A solution is to change the logical check to: `svd_rank >= 1 and isinstance(svd_rank, (int, np.integer))` in both the `rdmd.compute_rank` and `utils.compute_svd` functions.
  
**To Reproduce**
I use the example from Tutorials 2 and 14:
```
import numpy as np
from pydmd.bopdmd import BOPDMD
import pydmd


x1 = np.linspace(-3, 3, 80)
x2 = np.linspace(-3, 3, 80)
x1grid, x2grid = np.meshgrid(x1, x2)

# Set the random seed for reproducibility
seed = 1
# The noise amplitude matches the noise amplitude in tutorial 2.
sigma = 0.4

time_forecast = np.arange(6 + 0.4, 18 + 0.4, 0.4)
time = np.arange(0, 6 + 0.4, 0.4)
num_time_samples = len(time)
true_eigenvalues = (-.115 - 1j) * np.pi / 2

spatial_modes = 2 / np.cosh(x1grid) / np.cosh(x2grid) 
data_clean = spatial_modes[:, :, np.newaxis] * np.exp(true_eigenvalues * time)
snapshots_1d = data_clean.reshape(x1grid.shape[0] * x1grid.shape[1], num_time_samples)

optdmd = BOPDMD(svd_rank=0, use_proj=True)
optdmd.fit(snapshots_1d, time);

print('Actual shape of the projected basis:')
print(optdmd.proj_basis.shape)
print('Expected shape of the projected basis:')
print(pydmd.utils.compute_svd(snapshots_1d, svd_rank=0)[0].shape)
```

**Expected behavior**
The shape should have been (6400, 4).

**Additional context**
I will open a pull request fixing the issue.

PR #368, opened 2023-03-31
Int check bug

klapo:
This should close issue #367

### DIFF
--- a/pydmd/rdmd.py
+++ b/pydmd/rdmd.py
@@ -42,7 +42,7 @@ def compute_rank(X, svd_rank=0):
     elif 0 < svd_rank < 1:
         cumulative_energy = np.cumsum(s**2 / (s**2).sum())
         rank = np.searchsorted(cumulative_energy, svd_rank) + 1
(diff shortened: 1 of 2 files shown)


klapo:
So I was pretty silly to not run the tests locally, I apologize. It looks like the behavior detailed in issue #368 was used to inform the bopdmd tests. @sichinaga is the behavior noted in the issue intentional? Or do we need to take a quick look at the tests to update them?

sichinaga:
Hi @klapo! Thanks for catching this issue. This update does indeed call for updating the BOP-DMD tests. Specifically, it requires that the current test BOPDMD models have a fixed `svd_rank=2` rather than the default `svd_rank=0`. I'm updating the tests in my latest pull request #374!

Also note that in PR #374, I remedy issue #367 by casting the rank computed by `compute_rank` to an `int`, which is slightly different from your fix, but shouldn't conflict with your code at all. Thanks again for catcing this! :)

klapo:
Thanks for fixing the tests @sichinaga! I'll close this pull request.

PR #374, opened 2023-04-06
BOP-DMD RAM usage improvements

sichinaga:
Previously, in order to compute the mean and standard deviation statistics for modes, eigenvalues, and amplitudes, the `BOPDMD` module would store the resulting DMD diagontics at each bagging iteration and then compute the bagging statistics at the conclusion of all of the trials. As anticipated, this results in RAM usage issues when attempting to fit `BOPDMD` models to high-dimensional datasets.

In order to improve the memory usage efficiency of the `BOP-DMD` module, I've replaced this implementation with an iterative computation of the summations $\Sigma x_i$ and $\Sigma |x_i|^2$, both of which can later be used to estimate variance or standard deviation. This effectively removes the need to store the computed modes, eigenvalues, and amplitudes at each trial, hence significantly cutting down on memory usage. 

Additionally, I've added a `modes_std` attribute to the module for easy access to UQ metrics for the `BOPDMD` modes.


sichinaga:
Update: The last commit involves (1) casting the rank computed by `compute_rank` to `int` in order to fix the issue from #367 and (2) updates to the `BOPDMD` tests, as fixing this issue requires that the current test `BOPDMD` models have a fixed `svd_rank=2` rather than the default `svd_rank=0`.

Thanks again @klapo for catching issue #367! :)

fAndreuzzi:
Looks very good @sichinaga! Ready to merge for me

PR #416, opened 2023-06-29
Some more changes in readme

fandreuz:
Hi team, this is my proposal for the README. I removed some parts (e.g. citation is replaced by GitHub view) and in general I tried to make it more compact since it's still quite massive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
   </a>
 </p>
 <p align="center">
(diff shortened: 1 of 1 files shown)


mtezzele:
I personally like it. There is just a double instance for testing (probably left from a copy/paste)

fandreuz:
Thanks @mtezzele, fixed in 87f7c957822d91485495d31eb0202d336296c9e4

PR #431, opened 2023-07-18
BOP-DMD with eigenvalue constraints

sichinaga:
Hi everyone! 

Here's the BOP-DMD with constraints feature described in the [roadmap](https://github.com/PyDMD/PyDMD/wiki/Roadmap)! 

I've basically added a new parameter `eig_constraints` to the `BOPDMD` module, which allows users to provide constraints that they would like to enforce upon the eigenvalues computed via BOP-DMD. Currently implemented are options to constrain eigenvalues to the left-half plane or the imaginary axis. Another available option is to enforce that eigenvalues are present with their complex conjugate. Constraints may also be combined if the combination makes sense.

I've also provided a few small tests to make sure basic functionality and error catching is working.

Let me know what y'all think! :)


pydmd/bopdmd.py
```diff
@@ -59,6 +59,14 @@ class BOPDMDOperator(DMDOperator):
         previously-mentioned sorting methods is chosen depending on eigenvalue
         variance.
     :type eig_sort: {"real", "imag", "abs", "auto"}
+    :param eig_constraints: Set containing desired DMD operator eigenvalue
+        constraints. Currently available constraints are "stable", which
+        constrains eigenvalues to the left half of the complex plane,
+        "imag", which constrains eigenvalues to the imaginary axis, and
+        "conjugate_pairs", which enforces that eigenvalues are always
+        present with their complex conjugate. Note that constraints may be
+        combined if valid.
```
fandreuz, line 68: Just to make this look nicer in the docs:
```suggestion
    :param eig_constraints: Set containing desired DMD operator eigenvalue
        constraints. Currently available constraints are `"stable"`, which
        constrains eigenvalues to the left half of the complex plane,
        `"imag"`, which constrains eigenvalues to the imaginary axis, and
        `"conjugate_pairs"`, which enforces that eigenvalues are always
        present with their complex conjugate. Note that constraints may be
        combined if valid.
```

pydmd/bopdmd.py
```diff
@@ -107,6 +115,7 @@ def __init__(
         num_trials,
         trial_size,
         eig_sort,
+        eig_constraints,
```
fandreuz, line 118: Do you want to force users to set a specific constraint? Maybe you could assign an empty set by default (also to keep it retrocompatible).

pydmd/bopdmd.py
```diff
@@ -240,6 +250,48 @@ def _varpro_opts_warn(self):
...
+            num_pair = num_eigs // 2
+            new_eigs = []
+            # If given an odd number of eigenvalues, find the eigenvalue with
+            # the smallest imaginary part and take it to be a real eigenvalue.
+            if num_eigs % 2 == 1:
+                imag_magnitudes = np.abs(eigenvalues_sorted.imag)
+                ind_single = np.argmin(imag_magnitudes)
+                new_eigs.append(eigenvalues_sorted[ind_single].real)
+                eigenvalues_sorted = np.delete(eigenvalues_sorted, ind_single)
+            for i in range(num_pair):
```
fandreuz, line 277: I think this loop could be replaced by a more NumPy-friendly approach

pydmd/bopdmd.py
```diff
@@ -760,6 +813,14 @@ class BOPDMD(DMDBase):
         previously-mentioned sorting methods is chosen depending on eigenvalue
         variance. Default is "auto".
     :type eig_sort: {"real", "imag", "abs", "auto"}
+    :param eig_constraints: Set containing desired DMD operator eigenvalue
```
fandreuz, line 816: Same as the docs comment above

pydmd/bopdmd.py
```diff
@@ -798,6 +861,14 @@ def __init__(
         else:
             self._varpro_opts_dict = varpro_opts_dict
 
+        if eig_constraints is None:
+            self._eig_constraints = set()
+        elif not isinstance(eig_constraints, set):
+            raise ValueError("eig_constraints must be a set.")
+        else:
+            self._eig_constraints = eig_constraints
```
fandreuz, line 869:
```suggestion
        if eig_constraints is None:
            eig_constraints = set()
        elif not isinstance(eig_constraints, set):
            raise ValueError("eig_constraints must be a set.")
        self._eig_constraints = eig_constraints
```

pydmd/bopdmd.py
```diff
@@ -944,6 +1015,16 @@ def modes_std(self):
         """
         return self.operator.modes_std
 
+    @property
+    def eig_constraints(self):
```
fandreuz, line 1019: Do you think we need to expose this property as part of the public API? Maybe it's not needed (just asking, don't know the usecase)

fandreuz:
Thanks @sichinaga! Left some minor comments, I really like the set-based approach, very clean!

It looks like `black` didn't like your changes

pydmd/bopdmd.py
sichinaga: Done!

pydmd/bopdmd.py
sichinaga: Also done!

pydmd/bopdmd.py
sichinaga: I mostly went with this approach so that the code was as readable and intuitive as possible (since I have a hunch we may end up changing how we compute conjugate pairs in the future). Happy to change it up though!

pydmd/bopdmd.py
sichinaga: I mostly made this so that a user may recall the constraints passed to the model. The property doesn't really have many other uses other than this, so I'm open to cutting it if we feel having the property does more harm than good.

mtezzele:
@sichinaga this is just a minor fix: could you please fix the typo in the forecast method documentation (enssembles)? Since you are already editing the file...

pydmd/bopdmd.py
fandreuz: Okay okay let's keep it flexible. I'm resolving this comment

sichinaga:
@mtezzele the typo has been fixed!

@fandreuz Thank you for reviewing the code! I'm not really sure how to get the `black` test to pass, so any insight on that would be appreciated.

Let me know if there'a anything else I can do to improve the quality of this PR! :)

fandreuz:
> @fandreuz Thank you for reviewing the code! I'm not really sure how to get the `black` test to pass, so any insight on that would be appreciated.

@sichinaga  [Black](https://github.com/psf/black) is the formatter that we're using in PyDMD, you can install it with pip and reformat a file with `black fileName.py`

sichinaga:
> > @fandreuz Thank you for reviewing the code! I'm not really sure how to get the `black` test to pass, so any insight on that would be appreciated.
> 
> @sichinaga [Black](https://github.com/psf/black) is the formatter that we're using in PyDMD, you can install it with pip and reformat a file with `black fileName.py`

Done! The code is now reformatted :)

sichinaga:
Hi everyone! As an update, I made a few small edits to the tests that I wrote so that I could ressolve the Codacy issues being thrown. It also looks like the tests that this PR is currently failing are the ubuntu tutorial tests because of the `FileNotFoundError` error

```
FileNotFoundError: [Errno 2] No such file or directory: 'SegTrackv2/JPEGImages'
```

I'm quite unsure how to ressolve this since this branch should be up to date with the master PyDMD branch, though maybe these errors are simply minor. Let me know if I can make any further improvements! :)

sichinaga:
Oops sorry about that. I accidentally closed this PR because I was renaming some of my branches. The PR is back up though!

fandreuz:
@sichinaga actually I'm thinking we should disable the test on the problematic tutorial which is making the whole thing fal, it gives much more problems than added value...

sichinaga:
> @sichinaga actually I'm thinking we should disable the test on the problematic tutorial which is making the whole thing fal, it gives much more problems than added value...

Alright! Does that mean we're okay to merge? Or should we hold off for a bit?

I was just hoping to get this PR squared away as soon as possible because I found a pretty major bug inside of this file and I aim to fix it in a new PR while creating minimal merge conflicts.

PR #432, opened 2023-07-21
Fixed major variable projection bug in `BOPDMD`

sichinaga:
This update fixes a bug caught by Farbod Faraji, who brought this to our attention via email.

Basically due to the way that the `error` variable was being updated, the variable projection code would discard valuable results in the event that the algorithm needed to search for higher values of `_lambda`. I pressume that `BOPDMD` worked previously, but only in the event that this portion of the code either wasn't triggered, or was triggered after several iterations. This update fixes this issue in that now the algorithm is able to proceed if higher values of `_lambda` need to be found.

This PR also includes a few minor variable projection changes and fixes. Namely `pred_improvement` used to be a single-element array rather than a single-element, which would result in strange `_lambda` outputs (this is now fixed). I also replaced warnings with `print` statements for more concise and consistent readability.

I've also added flags called `converged` and `stalled` for readability, and because I'm preparing the module for an update where we take convergence into account when keeping or discarding BOP-DMD trials (since it doesn't make a lot of sense to keep the results from a trial of BOP-DMD if the trial didn't even converge).

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -477,6 +477,10 @@ class BOPDMDOperator(DMDOperator):
         B, residual, error = compute_residual(alpha)
         U, S, Vh = self._compute_irank_svd(Phi(alpha, t), tolrank)
 
+        # Initialize termination flags.
+        converged = False
+        stalled = False
+
         # Initialize storage.
         all_error = np.zeros(maxiter)
         djac_matrix = np.zeros((M * IS, IA), dtype="complex")
@@ -548,7 +552,7 @@ class BOPDMDOperator(DMDOperator):
                 0.5
                 * np.linalg.multi_dot(
                     [delta_0.conj().T, djac_matrix.conj().T, rhs_temp]
-                ).real
+                )[0].real
             )
             improvement_ratio = actual_improvement / pred_improvement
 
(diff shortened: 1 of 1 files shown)


fandreuz:
Looks good to me @sichinaga ! Are you able to reproduce the previous error wih a test to check that it does not happen anymore after your fix? If it's too complex no need to worry

sichinaga:
> Looks good to me @sichinaga ! Are you able to reproduce the previous error wih a test to check that it does not happen anymore after your fix? If it's too complex no need to worry

Sure @fandreuz !

Here's an example using the flow past a cylinder data found at dmdbook.com/DATA.zip.

```
import numpy as np
import scipy.io as sio
from pydmd import BOPDMD

X = sio.loadmat("CYLINDER_ALL.mat")["VORTALL"]
t = np.arange(X.shape[1])

bopdmd = BOPDMD(svd_rank=15, num_trials=0, varpro_opts_dict={"verbose":True})
bopdmd.fit(X, t)
```

The old code will result in a premature termination of the variable projection routine due to the issue described previously:
```
UserWarning: Failed to find appropriate step length at iteration 0. Current error 122139.88409300604.
```

The new code results in the following since it allows the variable projection routine to advance in the event that larger step lengths need to be found. In this case, variable projection continues until it terminates due to a stall, which is a separate termination condition.
```
Step 0 Error 116662.37823284771 Lambda 4096.0
Step 1 Error 116134.59636572568 Lambda 4820.921169452047
Step 2 Error 116134.59636572559 Lambda 40440817889.4348
Stall detected: error reduced by less than 1e-12 times the error at the previous step. Iteration 2. Current error 116134.59636572559.
```

sichinaga:
Hi @mtezzele !

Do you think we can possibly get this PR merged soon? This bug is just pretty fatal in my opinion and I think we should upload the fix asap. I'm happy to work on this PR some more though if there are additional edits we want to make to the `BOPDMD` module before merging! 😄

mtezzele:
@sichinaga Sure, I was waiting for Francesco but we can go ahead. Thanks and sorry for the delay.

mtezzele:
@sichinaga could you just rebase it please?

Issue #86, opened 2019-04-22
Support for separate X and Y inputs


abrazhe:
As far as I could understand, currently pydmd takes input as a single X matrix, from which the X and Y matrices are formed. 

Sometimes it is convenient to sample real data non-uniformly, e.g. in short batches  from the original dynamics. This would mean that the columns of X are not necessarily contiguous in time, but the corresponding X-Y pairs are. 

How about adding the option to the `DMD.fit` method to provide separate X and Y matrices?

mtezzele:
Hi! You are right, up to now only a single matrix is available. 
Feel free to implement the modification with some small tests, and then we can review the code together in a PR.

Kolaru:
Since I need this feature, I'm interested in implementing it. 

However I had a look at the source code and I'm a bit confused as to what to do with the stored snapshots and their shape (`DMDBase._snapshots` and `DMDBase._snapshots_shape`). Would it be okay to not have them in the general case and let some of the plotting functions (`DMDBase.plot_mode_2D` and `DMDBase.plot_snapshots_2D`) fail if the input was not a time series? (they seem to be the only place where the stored snapshots and their shape is used)

mtezzele:
@Kolaru sure, you can propose such a solution and then we will see together whether there is a better integration or not ;)
Contributions are more than welcome!

xlelephant:
Hi, @abrazhe @Kolaru ,

Any progress for the enhancement? I agree with the proposal of @abrazhe .

In dmdc, it is interesting that the `_fit_B_unknown(X, Y)` method of `dmdc` has already separated the X and Y inputs.
I guess that we only need to slightly change the related `fit()` functions to take the separated inputs.
So it could be something like:
`fit(self, X, I, B=None)` -> `fit(self, X, I, Y=None, B=None)
and if Y=None, use
        X = X[:,:-1]
        Y = X[:,1:]
else:
        X = X
        Y = Y

Kolaru:
I looked into it and for most of the method the change is indeed relatively straightforward. However for some, like `MrDMD` it seems to be either tricky or impossible to implement the separation.

I will try to open a PR with what I did so far soon.

kathryn-garside:
@Kolaru has there been any progress on creating the PR? I am also wanting to implement this for DMD/MrDMD to try and deal with missing data

Kolaru:
No unfortunately I moved to other projects, so I won't be able to pursue this contribution.

PR #436, opened 2023-07-29
DMD fitting with 2 input matrices

sichinaga:
Here is my first shot at incorporating the functionality discussed in Issue #86 and Discussion #411 into the `DMD` module!

Much like what was previously discussed, this PR introduces an optional argument to the `fit` method of the `DMD` class so that one may now fit their model with a call along the lines of the following. If the `Y` argument is not provided, the previous functionality is used, i.e. `Y` is taken to be the last $m-1$ snapshots of the provided $m$ data snapshots.
```
dmd.fit(X=data_1, Y=data_2)
```
The **most notable changes** that I made in order to accommodate for this new argument are summarized bellow:
- `DMDBase` now has the attribute `snapshots_holder_y`, which acts as storage for the optionally-provided $Y$ data matrix, and as a flag that indicates whether or not $Y$ data was provided during the fit. This attribute is appropriately set whenever $Y$ data is provided and is appropriately reset during calls to `fit`, much like the original `snapshots_holder` attribute, which is still used to house the `X` data.
- `DMDBase` now also contains the function `compare_data_shapes()`, which is used to ensure that the input data matrices $X$ and $Y$ are the same shape, if provided separately.
- The `optimal_dmd_matrices` function is (possibly the only major function) impacted by this change, in that whether someone provides $X$ and $Y$ or just $X$ changes how this function plays out. I did my best to use [the provided reference](https://polytechnique.hal.science/hal-00995141/document) in order to accommodate for all possible cases, and the code provided in this PR is what I came up with. I am by no means an expert on sparsity-promoting DMD however, so additional feedback on my solution would be appreciated.

Also! In the event that the provided $X, Y$ data has the shape $X,Y \in \mathbb{R}^{n \times m}$, my intended functionality is that `original_time` and `dmd_time` account for the $m$ time points $0, 1, \dots, m-1$. This is in contrast to fitting a model with only $X \in \mathbb{R}^{n \times m}$, in which case $X_1, X_2 \in \mathbb{R}^{n \times m-1}$ are used to fit the model and the time dicts account for the $m$ time points $0, 1, \dots, m-1$. This was a personal decision however, and feedback would be appreciated on this as well!

I've also added an elementary suite of tests for this new functionality. **Currently, I generally test for the following.** Happy to test for more things though!
- Error is thrown if $X$ and $Y$ are not the same shape.
- Time dictionaries are as described above.
- A `DMD` model fitted with `fit(X=data)` is the same as a `DMD` model fitted with `fit(X=data[:, :-1], Y=data[:, 1:])`.
- A `DMD` model fitted with `fit(X=data, Y=shifted_by_dt(data))` is accurate.
- A `DMD` model fitted with `fit(X=data, Y=derivative(data))` is accurate.

The third bullet point above is tested for all combinations of the `exact` and `opt` parameter in order to assess the continued functionality of the `optimal_dmd_matrices` function. It's also worth noting that time dict and `snapshots` differences make it so that models are only close, not equal, when `opt=True, exact=False` for the third bullet point (this is reflected in the `test_equal_models_opt` test).

**Other things I've considered, but haven't necessarily incorporated into this PR yet:**
- The impact on plotting utilities. Honestly, I don't really think much will be impacted. Just as it stands now, if someone fits their model with two data matrices $X,Y$, our current snapshot plotter will only consider the $X$ snapshots. I suppose we could implement another feature that allows users to plot both $X$ and $Y$ if provided, though maybe that's a PR for another day...
- The impact on other PyDMD modules. I acknowledge that this functionality may be desirable in other modules, however for some like `BOPDMD`, it actually doesn't make sense. This PR only implements this feature for the `DMD` class and thus it impacts its children who use the `DMD` fit method. I haven't done much to edit/alter the code for said children though... (and I don't really think it's necessary at the moment unless I'm mistaken)

Apologies for the lengthy message. I mostly just wanted to get all of my current work and thoughts out there asap so that we can potentially start collaborating on this as a team/community, because I really think that this feature would be very valuable to the PyDMD package. I'm also more than willing to take the time to ensure that this feature is done well. Any and all help with this feature is greatly appreciated. :octocat:😄


pydmd/dmd.py
```diff
@@ -46,20 +46,28 @@ class DMD(DMDBase):
...
         n_samples = self.snapshots.shape[1]
-        X = self.snapshots[:, :-1]
-        Y = self.snapshots[:, 1:]
+
+        if Y is None:
+            X = self.snapshots[:, :-1]
+            Y = self.snapshots[:, 1:]
+        else:
+            self._snapshots_holder_y = Snapshots(Y)
+            self._compare_data_shapes()
```
fandreuz, line 68: I would prefer the check to happen *before* `_snapshot_holder_y` is assigned. `_compare_data_shapes` could take `y` as a parameter.

pydmd/dmdbase.py
```diff
@@ -724,6 +744,16 @@ def _compute_amplitudes(self):
...
 
+    def _compare_data_shapes(self):
+        """
+        Method that ensures that the data inputs X and Y are the same shape
+        if provided separately. Throws an error if the shapes do not agree.
+        """
+        if (self._snapshots_holder and self._snapshots_holder_y) and (
+            self.snapshots.shape != self.snapshots_y.shape
+        ):
+            raise ValueError("X and Y input data must be the same shape.")
```
fandreuz, line 755:
```suggestion
            raise ValueError(f"X {X.shape} and Y {Y.shape} input data must be the same shape.")
```

pydmd/dmdbase.py
```diff
@@ -672,13 +686,19 @@ def _optimal_dmd_matrices(self):
...
+
+            _, s, V = compute_svd(X, self.modes.shape[-1])
+            n = X.shape[1]
 
             q = np.conj(
                 np.diag(
                     np.linalg.multi_dot(
                         [
-                            vander[:, :-1],
+                            vander[:, :n],
```
fandreuz, line 701: `n` is only used once:
```suggestion
                            vander[:, :X.shape[1]],
```

If you apply this, please delete the definition of `n` above.

fandreuz (commented):
Thanks @sichinaga! I've left some minor comments, your tests are more than enough for me.

Just an implementation-related question, have you considered making the `X,Y` approach the default one in `DMDBase` internals? I'm not a huge fan of flags, especially in a huge class like `DMDBase`. We could just make sure DMD subclasses always set `X` and `Y` (`Y` being `X[:, 1:]` if the user passes `None`), thus removing the need for all the `if/else`s.

pydmd/dmd.py
sichinaga: Gotcha! I just updated `self._compare_data_shapes()` so that it accepts the $Y$ snapshots and compares their shape against the `self.snapshots_shape` property.

pydmd/dmdbase.py
sichinaga: Done!

sichinaga:
> Just an implementation-related question, have you considered making the `X,Y` approach the default one in `DMDBase` internals? I'm not a huge fan of flags, especially in a huge class like `DMDBase`. We could just make sure DMD subclasses always set `X` and `Y` (`Y` being `X[:, 1:]` if the user passes `None`), thus removing the need for all the `if/else`s.

@fandreuz I have considered this! However, I opted for the current implementation for a couple of reasons.

So unless I'm mistaken, the `snapshots` parameter was only ever previously used for a couple of things beyond declaring the `X` and `Y` matrices that are used for operator computations in the `fit` methods: (1) for computing amplitudes, such as in the `optimal_dmd_matrices` function, and (2) for viewing/plotting so that users can recall the snapshots that were used to generate their DMD model. As such...

- The number of occurrences of the `if/else` flag is pretty minimal, and it only ever really comes up in the `optimal_dmd_matrices` function which is triggered when `opt=True`.
- If one of the main purposes for `snapshots` is plotting and reference, it's a bit redundant to store both `X` and `Y` if `snapshots=X[:, :-1]` and `snapshot_y=X[:, 1:]`. This is not necessarily true if $Y$ is provided separately, in which case it makes sense to store both `snapshots` and `snapshots_y`.

The availability of `snapshots_y` now makes it so that this $Y$ data-fitting functionality is available if we ever want to add the `Y` data-fitting parameter to other modules. It sits idly and everything functions as before otherwise. On top of reducing the invasiveness of this feature, this also makes sense considering how providing a custom $Y$ data matrix doesn't make sense for some DMD methods.

sichinaga:
Thanks for the feedback @fandreuz ! I believe I addressed all of your comments, though do let me know if you would change anything else! :)

Some additional technical things:
- Codacy seems to be upset with my testing code for docstring reasons, not liking my variables, etc. How concerned should I be about this? I'm happy to clean up the testing code, transfer the toy data creation code to another file, etc.
- I introduced the function `differentiate` to the testing file, which performs 2nd order finite difference. Do you think this might be a function we want available in another file such as `utils.py` or even some sort of file for data preprocessor functions?

pydmd/dmd.py
fandreuz: Perfect, thanks!

fandreuz:
> Long story short, I think it would suffice to make X,Y fitting the default behavior for `DMD`, but not necessarily for `DMDBase`.

Thanks for the explanation @sichinaga, looks good indeed :+1: 

> Codacy seems to be upset with my testing code for docstring reasons

In general we're quite flexible with Codacy since it reports many issues which we can safely ignore. However in this case I see they are quite meaningful (and easy to fix), since it's better to avoid overriding global variable names with function parameter names. And the same for the unused argument. You can definitely ignore the one about the docstring.

> Do you think this might be a function we want available in another file such as utils.py or even some sort of file for data preprocessor functions?

I propose for now we keep it where it is, and the first time it's used somewhere else we will move it.

sichinaga:
Sounds good @fandreuz !

I've fixed all of the Codacy issues with the exception of 2, one of them being a minor "missing module docstring" issue (which we can probably ignore), and an "unused argument" issue, which I don't think I can really resolve because the function with the unused argument is the function that defines the autonomous ODE system that I'm using to generate testing data, and the function needs to be of that format in order to be passed to `solve_ivp` (documentation [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html)).

Basically, I think that this last commit should fix all major and all fixable Codacy issues! Let me know if there's anything else I can do to prepare this PR for merging! :)

fandreuz:
Looks good for me! Thanks @sichinaga

mtezzele:
I proceed with the merge. Let's not forget about showing this new features in one of the tutorial, even one of the first ones is ok, by explicitly passing the two matrices. Thanks @sichinaga for the nice work!

mtezzele:
@sichinaga the automatic rebase does not work (no idea why). Could you rebase from your side please?

sichinaga:
> @sichinaga the automatic rebase does not work (no idea why). Could you rebase from your side please?

Just rebased!

PR #444, opened 2023-08-10
Custom input functions for BOP-DMD eigenvalue constraints

sichinaga:
Hi everyone!

This will probably be the last major BOP-DMD eigenvalue constraint feature update in a while, but since some users seem to occassionally be getting strange results when using the built in functions for constraining eigenvalues in the `BOPDMD` module, I've decided to add more customization options by allowing users to define and input their own eigenvalue constraint functions.

Basically, the previous functionality was that if users wanted to constrain their eigenvalues when applying BOP-DMD, they could provide a set of desired eigenvalue constraints via the `eig_constraints` parameter. The arguments provided would then activate a series of built in functions that would constrain the BOP-DMD eigenvalues at each iteration of the variable projection routine depending on the arguments.
```
bopdmd = BOPDMD(eig_constraints={"stable", "conjugate_pairs"})
```

This PR introduces the possibility of manually defining and inputting a custom function to be applied at each iteration of the variable projection routine. This way, users have the freedom to develop and input their own eigenvalue constraint functions in the event that they are either unhappy with the built in functions, or would like to enforce some sort of new type of eigenvalue constraint.
```
bopdmd = BOPDMD(eig_constraints=my_function)
```

In this PR, I include a variety of tests that ensure that `eig_constraints` can not only be a function, but it must also be a function that specifically takes in a single n-dimensional array and outputs a single n-dimensional array. I also test the functionality of the module for a small handful of cases. (Also included in this PR is a revision to my original ODE-solving code for better readability via the `solve_ivp` function, but this is more of a minor change.)

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -1082,7 +1082,7 @@ class BOPDMD(DMDBase):
         if isfunction(eig_constraints):
             if self._svd_rank >= 1 and isinstance(self._svd_rank, int):
                 r = self._svd_rank
-            else: # use a dummy rank of 10
+            else:  # use a dummy rank of 10
                 r = 10
 
             try:
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -13,6 +13,7 @@ temporal uncertainty-quantification. 2021. arXiv:2107.10878.
 
 import warnings
 from collections import OrderedDict
+from inspect import isfunction
 from scipy.sparse import csr_matrix
 from scipy.linalg import qr
 import numpy as np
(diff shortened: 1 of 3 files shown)


sichinaga:
I would additionally like to note that the only codacy issue being thrown by this PR (the "unused argument" issue) is *not* fixable due to the fact that the function throwing this error must have this particular function header in order to be compatible with the `solve_ivp` function ([documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html)).

Basically this PR should be up to standards, but I am of course open to making any necessary edits or changes before merging. :)

Issue #283, opened 2022-06-09
No tutorial for `HankelDMD`


fAndreuzzi:
We have a tutorial for `HODMD`, but not for `HankelDMD` (which is a simpler variant of DMD wrt `HODMD`).

PR #445, opened 2023-08-11
`PrePostProcessingDMD`

fandreuz:
As discussed during the last PyDMD meeting, I've implemented a simple DMD decorator with pre/post processing support.


fandreuz:
@sichinaga for the tutorial you're preparing it would be cool to showcase how to make a pipeline of preprocessings, is there any other standard pre-processing you think could be implemented?

sichinaga:
> @sichinaga for the tutorial you're preparing it would be cool to showcase how to make a pipeline of pre-processings, is there any other standard pre-processing you think could be implemented?

Agreed! I'm not sure if some of these count as preprocessing steps, but I'll just throw some thoughts out there:
- Checking for rank deficiency, which would simply involve comparing the rank of the data against the rank of a time-delay matrix built using the data (as discussed in our meeting).
- Tools for computing data derivatives, as some users may want to apply DMD in order to compute a continuous-time operator $x'=Ax$. The preprocessor would probably do something along the lines of compute a derivative matrix via finite difference, fit a model using $X, \dot{X}$, and then have to remember that a continuous-time model was computed rather than a discrete-time one?

(Also things that are currently implemented in PyDMD that might actually do well as preprocessing options):
- Converting the data to a hankel matrix before applying DMD, de-hankeling reconstructions afterwards (this can already be done by simply applying `HankelDMD`).
- Checking data for high condition numbers (done in Snapshots I believe).

fandreuz:
> Converting the data to a hankel matrix before applying DMD

This looks very cool, but unfortunately it does not fit the current interface of `PrePostProcessingDMD`. The modification to the input matrix must be in-place, and the Hankel matrix is larger than the original one.

The in-place requirement is a little bit constraining but I feel it helps in keeping the thingy much simpler, since we are allowed to dedicate the whole output of the pre-processing to the state retained to be passed as part of the post-processing input (e.g. see `mean` in the implementation above).

Some options:
- We could provide different versions of the `PrePostProcessing`, for instance the one I implemented could be renamed to `StatefulInPlacePrePostProcessing`, and we could have another one called `StatelessPrePostProcessing` which could implement the `Hankel` transformation
- We build a more complicated `PrePostProcessing` which requires a `pre_processing(state_holder, *args, **kwargs)` which is responsible for putting some state in `state_holder` (to be later passed to `post_processing`) and to return a tuple of things to pass to `dmd.fit()`.

sichinaga:
> * We build a more complicated `PrePostProcessing` which requires a `pre_processing(state_holder, *args, **kwargs)` which is responsible for putting some state in `state_holder` (to be later passed to `post_processing`) and to return a tuple of things to pass to `dmd.fit()`.

I think the most ideal thing to do if possible is to design `PrePostProcessing` so that it can deal with as many pre and post processing situations as possible, including both data centering and time-delay building. My main reservation is that there are very few additional in-place pre and post processing pipelines that I can think of, so I'm not sure how useful separating the modules would be...

But with that being said, another potentially useful preprocessor I just thought of is some kind of data smoothing technique such as kernel smoothing. This modification would also be in-place if I'm not mistaken!

Basically I'm not sure what would be the best option here, but I personally think that (1) centering and (2) time-delay would be genuinely useful and fairly popular out of all of the preprocessors discussed so far. Open to more input and feedback from the rest of the group though!

fandreuz:
@sichinaga I've re-implemented `PrePostProcessingDMD` to allow non in-place operations! We also have Hankel-preprocessing now

fandreuz:
Updated summary of the changes:
- `PrePostProcessingDMD`
- Zero-mean pre-processing
- Hankel pre-processing
- SVD projection pre-processing
- Re-implemented `HankelDMD` with Hankel pre-processing
- Re-implemented `HODMD` with SVD projection pre-processing
- `HAVOK` needed a few changes because it cannot be anymore a subclass of `HankelDMD`

sichinaga:
Wow, thank you so much @fandreuz this is so thorough! :D

I haven't gotten around to fully reviewing all of the files just yet so this might be a silly question, but with this current framework, would it be fairly straight-forward for users to, say combine pre and post processing steps? Like for example, if a user wanted to first hankel their data and then apply centering before applying a DMD fit (in which case, the post processing would probably involve first re-applying the averages and then de-hankeling), what kind of user input / syntax would that involve?

fandreuz:
Ready to merge for me, any comment? @sichinaga @ndem0 @mtezzele

sichinaga:
> Ready to merge for me, any comment?

Sorry @fandreuz this comment might have gotten lost in all of the updates, but I'll just bring it up again here:

In the original DMD with centering paper, a mean is computed for each row of the data matrix $X$, with each row of data then having its row average subtracted off.

$$
X_2 = A X_1 \quad \Longrightarrow \quad \overline{X}_2 = A \overline{X}_1
$$

$$
\overline{X} = X - \mu 1^\top \qquad \mu = \frac{X 1}{1^\top 1}
$$

Basically, all the current centering code needs is to apply `np.mean` along `axis=1` and then perform the appropriate subtraction, like so.
```
def center(X):
    row_avgs = np.average(X, axis=1)
    avg_mat = np.tile(row_avgs, (X.shape[1], 1)).T
    X_centered = X - avg_mat
    return X_centered, avg_mat
```

Other than that, I think the code looks good! I'll give it a final thorough pass right now though!

sichinaga:
Thanks again for making this tool @fandreuz ! This is awesome!

After this is merged, I'll make a new tutorial that demonstrates centering, hankeling, and how one might combine them or alternatively opt for using `HankelDMD`. This should end up addressing issue #283 too.

We also might want to think about making a separate developer tutorial on how one might create a custom pre/post processor?

fandreuz:
Hi @sichinaga, I've completely missed your comment. The behavior is rectified in the last commit, you can pass the parameters `axis` and `weights` to the preprocessing initiator. By default `axis=1`.

sichinaga:
> Hi @sichinaga, I've completely missed your comment. The behavior is rectified in the last commit, you can pass the parameters `axis` and `weights` to the preprocessing initiator. By default `axis=1`.

Awesome! Thanks @fandreuz :)

sichinaga (approved):
Everything looks okay to me!

PR #450, opened 2023-08-18
`BOPDMD` updates: discard bad trials and fix eigenvalue sorting

sichinaga:
Hi all!

So one common issue with `BOPDMD` as it is now is that when users wish to perform `num_trials` many iterations of variable projection (i.e. the "bagging"part of bagging, optimized DMD), the current code performs exactly `num_trials` many iterations. However sometimes, variable projection fails to converge, and when/if that happens,  it makes a lot more sense to discard the results from said trial since the results can be quite inaccurate and hence mess up the running averages of the method.

This PR aims to mitigate this issue by keeping track of whether or not variable projection converged via a returned `converged` flag and by discarding results from trials that don't converge. I've also added the parameters `bag_warning` and `bag_maxfail` so that users can control when/if they receive warnings or failures after a certain number of consecutive variable projection failures. There are a few additional minor edits in this PR, but discarding bad bags is the main feature I'm focused on here.

I haven't created any tests for this feature yet since it won't necessarily be easy to get a data set with a known/deterministic amount of failed variable projection trials. Though if anyone has any ideas for a good test for this functionality, please let me know! I'll probably be testing this functionality locally in the meantime!


pydmd/bopdmd.py
```diff
@@ -432,10 +482,11 @@ def _variable_projection(self, H, t, init_alpha, Phi, dPhi):
         :param dPhi: (M, N) matrix-valued function dPhi(alpha,t,i) that
             contains the derivatives of Phi wrt the ith component of alpha.
         :type dPhi: function
-        :return: Tuple of two numpy arrays representing...
+        :return: Tuple of two numpy arrays and a boolean representing...
```
fandreuz, line 485:
```suggestion
        :return: Tuple of two numpy arrays and a boolean representing:
```

fandreuz (commented):
Thanks @sichinaga :+1:

sichinaga:
Thank you @fandreuz for the feedback! I've fixed all of the issues and the code should reflect your suggestions. 😄 

@fandreuz @mtezzele @ndem0 
Apologies that this PR is getting bigger, but while editing the `BOPDMD` code, I caught a bug and made a few additional adjustments, which I will summarize here. This list excludes minor documentation edits, naming edits, etc.
- When I introduced the eigenvalue constraining functionality, I introduced an eigenvalue sort, which I now realize could erroneously scramble the eigenvalues. The sorting is now removed.
- During the variable projection routine, the computed eigenvalues naturally come out as complex conjugate pairs (for example, something like: `[-0.1+1j, -0.1-1j, 0.51+0.1j, 0.49-0.1j, 1.3+0.0001j]`). I realized that I wasn't taking advantage of this structure previously when users requested the `conjugate_pairs` constraint. This is now implemented.
- The introduction of bag removal makes things a bit tricky if one's error tolerance is too low. Hence in order to make small tolerances more realistic to reach, I've introduced relative error as the go-to proxy for variable projection termination.

ndem0:
We have only one error using Win and Py3.7, but it looks just matter of the tolerance in the assert function. I would try to increase the relative one to 0.002 (no idea about the reason though).
Except for that, green light also for me, thanks Sara!

sichinaga:
> We have only one error using Win and Py3.7, but it looks just matter of the tolerance in the assert function. I would try to increase the relative one to 0.002 (no idea about the reason though). Except for that, green light also for me, thanks Sara!

Thanks @ndem0 !

I just increased the relative tolerance to 0.01 for some extra buffer. I noticed that this test occasionally fails, probably due to some amount of randomness in the bagging routine. I also added an extra test for thoroughness, though it looks like it already failed once, again due to some amount of randomness. Re-running the test should probably do the trick though!

I also noticed that the tutorial tests were failing a couple of commits ago, though I'm not entirely sure why...

mtezzele:
@sichinaga I think it is sufficient to increase the `atol` for a couple of `assert_allclose` and the tests will pass.

@fandreuz I stopped the tutorial_test actions because after 5 hours they were still running. Is there something we can do to simplify it? Probably it is related with the discussion #456 ...

sichinaga:
> @sichinaga I think it is sufficient to increase the `atol` for a couple of `assert_allclose` and the tests will pass.
> 
> @fandreuz I stopped the tutorial_test actions because after 5 hours they were still running. Is there something we can do to simplify it? Probably it is related with the discussion #456 ...

Thanks @mtezzele, I'll be sure to update the tests, though probably after we get the tutorial testing issues squared away!

mtezzele:
@sichinaga let's try with a rebase to see if with the last two PRs the situation has improved.

sichinaga:
Thanks @mtezzele and @fandreuz !

I loosened the tests up a bit, and I believe that the functionality that I introduced in this PR was triggering an excessively long loop in the `BOPDMD` tutorial, which should be corrected for now. Sorry for the trouble that this may have caused, but I think everything should be working now! 😅 

Ready to merge when you are. :)

PR #458, opened 2023-08-26
Timeout notebooks run and export when running for too long

fandreuz:
Should patch the problem reported by @mtezzele under #450. `nbconvert` is a black box and there's not much room to improve things in a more reliable way unfortunately. When it hangs I think the best we can do is to re-trigger the pipeline


mtezzele:
Thanks. This is ok for me. Merge when you want.

PR #465, opened 2023-10-09
New module: LANDO (plus minor updates)

sichinaga:
Hi all!

Moving on with additions from the [roadmap](https://github.com/PyDMD/PyDMD/wiki/Roadmap), here is the `LANDO` module, which follows the work of [this paper](https://royalsocietypublishing.org/doi/10.1098/rspa.2021.0830).

In short, the linear and nonlinear disambiguation optimization (LANDO) algorithm is a kernelized generalization of DMD that first uses kernel methods to solve for an approximate function $f(\cdot)$ such that

$$
\mathbf{y} \approx f(\mathbf{x})
$$

using snapshots from $\mathbf{X}, \mathbf{Y} \in \mathbb{R}^{n \times m}$. Baddoo et al. then show that this expression may be decomposed into a bias term, linear expression, and nonlinear expression about a fixed point $\bar{\mathbf{x}} = \mathbf{x}-\mathbf{x}'$, from which one may obtain modes, eigenvalues, etc.

$$
f(\mathbf{x}) = \mathbf{c} + \mathbf{L} \mathbf{x}' + \mathbf{N}(\mathbf{x}')
$$

The `LANDO` module provides the tools necessary for performing both fitting steps, while also allowing users to access all interesting outputs, such as the `f()` function, the `bias`, `linear`, and `nonlinear()` terms, the `modes` and `eigs` of $\mathbf{L}$, etc. The module also supports a variety of functionalities and customization options, including on-the-fly updates, customizable kernel functions, system integration, data rescaling, and more. I've basically ported over all functionalities of the [original LANDO code](https://github.com/baddoo/LANDO) while also incorporating in a few new features and overall compatibility with the PyDMD package.

This PR also contains...

- a basic testing suite that examines the Lorenz system
- a tutorial that walks through all of the major parameters and features of the module, and
- a few miscellaneous updates (swapping the axes of one of the `plot_summary()` plots, making the `BOPDMD` verbose option more readable when there are many bags, etc.)

Apologies for the delay on this module -- it kind of took me a while to implement all of the functionalities that I thought were actually user-friendly and practical. As such, I know that this is quite a lengthy PR, so no need to rush too much with the review. I would love to hear everyone's thoughts and opinions though! 😃 

Kind regards,
Sara Ichinaga


pydmd/bopdmd.py
```diff
@@ -748,7 +748,7 @@ def compute_operator(self, H, t):
                 "Consider re-adjusting your variable projection parameters "
                 "with the varpro_opts_dict and consider setting verbose=True."
             )
-            warnings.warn(msg)
```
fandreuz, line 751: Is there a reson to prefer `print` over `warnings.warn`? The latter can be disabled by the user

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+import numpy as np
+from scipy.integrate import solve_ivp
+from sklearn.metrics.pairwise import pairwise_kernels
+
+from .dmdbase import DMDBase
+from .dmdoperator import DMDOperator
+from .snapshots import Snapshots
+from .utils import compute_svd, compute_tlsq
+
+SUPPORTED_KERNELS = ["linear", "poly", "rbf"]
```
fandreuz, line 23: I propose to transform this to an `enum`. Or at least a tuple

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+        :type kernel_function: function
+        """
+        # Determine the order in which to parse the data snapshots.
+        parsing_inds = np.arange(X.shape[1])
+        if self._permute:
+            np.random.shuffle(parsing_inds)
+
+        # Initialize the Cholesky factorization routine.
+        ind_0 = parsing_inds[0]
+        x_0 = np.expand_dims(X[:, ind_0], axis=-1)
```
fandreuz, line 181: I think it's more clear this way:
```suggestion
        x_0 = X[:, ind_0][..., None]
```

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+
+        # Initialize the online learning routine, if applicable.
+        if self._online:
+            self._cholesky = C
+            self._P = np.ones((1, 1))
+            self._weights = y_0 / (self._cholesky[0][0] ** 2)
+
+        for ind_t in parsing_inds[1:]:
+            # Grab the next corresponding pair of snapshots.
+            x_t = np.expand_dims(X[:, ind_t], axis=-1)
```
fandreuz, line 194: Same as above

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+            # Get the results of this Cholesky factorization iteration.
+            results = self._cholesky_step(x_t, kernel_function, self._cholesky)
+            _, s_t, _, k_tt, delta_t = results
+
+            # NOT almost linearly dependent - update dict and cholesky factor.
+            if np.abs(delta_t) > self._dict_tol:
+                self._sparse_dictionary = np.hstack(
+                    [self._sparse_dictionary, x_t]
+                )
+                self._cholesky = np.vstack(
```
fandreuz, line 275: Is there any way to reduce duplication between `update_operator` and `compute_operator`?

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+                    "by setting the kernel_gradient parameter the LANDO model."
+                )
+                raise ValueError(msg)
+            return self._kernel_gradient(X, y)
+
+        if self._kernel_metric == "linear":
+            return X.T
+
+        # Kernel metric must be polynomial or RBF.
+        if "gamma" in self._kernel_params.keys():
```
fandreuz, line 586: `.keys()` is not needed:
```suggestion
        if "gamma" in self._kernel_params:
```

fandreuz: Same for other occurrences below

pydmd/lando.py
```diff
@@ -0,0 +1,1089 @@
...
+        # Note: the attribute operator.sparse_dictionary is re-scaled.
+        # Thus we undo this, but only when the user asks for the dictionary.
+        if not self.partially_fitted:
+            raise ValueError("You need to call fit().")
+        if isinstance(self._x_rescale, np.ndarray):
+            return np.divide(
+                self.operator.sparse_dictionary,
+                np.expand_dims(self._x_rescale, axis=-1),
+            )
+        return (1 / self._x_rescale) * self.operator.sparse_dictionary
```
fandreuz, line 644:
```suggestion
        return self.operator.sparse_dictionary / self._x_rescale
```

fandreuz (commented):
Thanks @sichinaga! I left some comments

ndem0:
Thanks Sara! Everything is cool, in addition to Francesco's comments I have only some concerns related to the kernel arguments. At the moment `LANDO` takes 4 optional inputs for the kernel
```
        kernel_metric="linear",
        kernel_params=None,
        kernel_function=None,
        kernel_gradient=None,
```

Sorry if I didn't understand something, but `kernel_function` should be just a custom "metrics" the user can pass right? Maybe we can consider joining these arguments in a unique dictionary the user can manipulate.

pydmd/lando.py
ndem0: I also propose to move the line inside the `LANDO` class (less overloading problems, more robust and safe), maybe like a private attribute `_supported_kernels` and the corresponding property to give the user easy access to the options directly from the code

pydmd/bopdmd.py
sichinaga: I mostly changed this because when you run `BOPDMD` with `verbose=True`, the output can get very lengthy. Also with all of the trials being run, there's a lot of potential for this particular warning to go up multiple times, and it just gets pretty chaotic and difficult to read. The print statements allow the output to be lengthy but also more clean and easy to read, even if some of the trials fail.

I hope this is okay? I can always put the warning back in, but for this reason, I decided to save warning messages for other specific issues.

sichinaga:
> Thanks Sara! Everything is cool, in addition to Francesco's comments I have only some concerns related to the kernel arguments. At the moment `LANDO` takes 4 optional inputs for the kernel
> 
> ```
>         kernel_metric="linear",
>         kernel_params=None,
>         kernel_function=None,
>         kernel_gradient=None,
> ```
> 
> Sorry if I didn't understand something, but `kernel_function` should be just a custom "metrics" the user can pass right? Maybe we can consider joining these arguments in a unique dictionary the user can manipulate.

@ndem0 Thanks for reviewing! 

No worries! I admit the setup is quite confusing. 😅 So `kernel_metric` and `kernel_params` work together and define the arguments that get passed to [`sklearn.metrics.pairwise.pairwise_kernels()`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.pairwise_kernels.html). This is one way users can define their kernel function of choice. However, since using this function limits users to the kernel functions defined in sklearn, I introduced `kernel_function` and `kernel_gradient` so that users have the option to define custom kernel functions that, when provided, override the `kernel_metric` and `kernel_params` arguments.

Do you still think we should alter these parameters?

pydmd/lando.py
sichinaga: Done! Changed this into a tuple that's saved under the `LANDO` attribute `_supported_kernels`, which can be accessed via the `supported_kernels` property of the `LANDO` class.

pydmd/lando.py
sichinaga: Done! They should all be changed now.

pydmd/lando.py
sichinaga: Fixed!

pydmd/lando.py
sichinaga: Fixed!

pydmd/lando.py
sichinaga: Yeahhh I did my best to consolidate these methods, but I will try to cook up some more solutions...

pydmd/bopdmd.py
fandreuz: I'll just throw this here since it's something I would like to do eventually in PyDMD: https://docs.python.org/3/library/logging.html

Basically the `logging` module allows you to have multiple levels (error, warning, info, debug) which the user may enable or disable (e.g. show only logs more important than `info` level). This way you can just pile as much output as you want, and the user is free to decide what they want to see.

If you feel it would simplify output management in this MR you could give it a try, otherwise let's work on this in another MR @sichinaga

pydmd/lando.py
fandreuz: Thanks!

pydmd/bopdmd.py
sichinaga: Hmm... since this part of the pull request is separate from the `LANDO` module (I just noticed that the `BOPDMD` verbose output looked terrible if multiple trials were being used, so I went with this very quick fix), I would like to incorporate this into a different PR if that's alright with you @fandreuz 

Thanks for sharing this though! This looks like a great resource. I will definitely be thinking about how to incorporate it into my modules.

sichinaga:
Okay so I think that just about wraps up most of the comments.

Thanks again @fandreuz and @ndem0 for reviewing the code! I may need to relax some of my tests, but overall, I think I'm just about done with this module! I also re-ran the tutorial for this module and everything seems to be working as expected. The other thing is that Codacy seems to be a bit upset with some of my code, but all of the medium priority issues are related to some `solve_ivp` syntax things that I can't really resolve.

Let me know if there are any additional things that I should address before merging. :)

sichinaga:
@mtezzele @fandreuz @ndem0 Just pinging this PR again because I think it's ready to merge!

PR #466, opened 2023-10-12
Fix missing submodules when using `pip install .`

xmjiao:
Add find_packages() to setup.py

This change adds `find_packages()` to the `setup.py` file, which will include all subpackages in the installation process. This ensures that the submodules (such as `preprocessing`) are included when installing the project using `pip install .`.


fandreuz:
Hi @xmjiao, thanks for the PR. Good to merge for me

PR #469, opened 2023-10-27
Revisiting Variable Projection for DMD.

greinerth:
The variable projection for DMD is reformulated to be compatible with SciPy's nonlinear least squares optimizer.
The optimization can be accelerated by at least a factor of two. Confer experimental setup [here](https://github.com/greinerth/pydmdstats).


pydmd/varprodmd.py
```diff
@@ -0,0 +1,721 @@
...
+        tau_star = median * omega
+
+    rank = np.where(sigma_svd >= tau_star)[0]
+    r_out = 0
+    if rank.size == 0:
+        r_out = sigma_svd.shape[-1]
+
+    else:
+        r_out = rank[-1] + 1
+    return r_out
```
mtezzele, line 86: I think this can be rewritten as:

```
if rank.size == 0:
    return sigma_svd.shape[-1]
return rank[-1] + 1
```

mtezzele (commented):
Hi, thank you for the contribution. Could you please change the documentation style to the one we are using?

sichinaga:
Thank you @greinerth for the module!

I would like to point out that the `BOPDMD` module uses the same reference, as Optimized DMD and bagging, optimized DMD (BOP-DMD) use variable projection behind the scenes, though the `BOPDMD` module is based more on the [original Optimized DMD](https://github.com/duqbo/optdmd) code and it doesn't currently utilize the SciPy nonlinear least squares optimizer.

@mtezzele @fandreuz @ndem0 Should we think of a way of combining the two scripts / approaches to Optimized DMD to avoid redundancy amongst the modules?

pydmd/varprodmd.py
greinerth: Will do, I will be able to do so on Monday. ;)

mtezzele:
@fandreuz @ndem0 @sichinaga @greinerth we should decide what are the possible options in terms of integration of this module and BOP-DMD.
@greinerth could you please help us by emphasizing the main characteristics of this module that are not covered by the BOP?

greinerth:
One characteristic is that SciPy's  [nonlinear least squares optimizer](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html) is used, which in many cases accelerates the optimization by a factor of two when performing pure variable projection. That means that BOP-DMD trains one model on the data. Also, a library selection scheme (QR decomposition with column pivoting) is integrated to preselect samples, which also can accelerate the computations.

My suggestions:
- Check if the implemented operator (VarProOperator) can be used in BOP-DMD.
- Check if the library selection can also be useful for BOP.

Please let me know if I can support you.

greinerth:
For additional clarification: SciPy's optimizer does not support complex numbers. Thus, the whole problem is transformed into real numbers. This generally affects the calculation of a Jacobian matrix and a residual, which are necessary for the Levenberg-Marquardt algorithm.

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+from typing import Any, Dict, Tuple, Union
+import numpy as np
+from scipy.optimize import OptimizeResult, least_squares
+from scipy.linalg import qr
+from .dmd import DMDBase
+from .dmdoperator import DMDOperator
+from .snapshots import Snapshots
+
+
+OPT_DEF_ARGS: Dict[str, Any] = {  # pylint: disable=unused-variable
```
fandreuz, line 29: This should be immutable, so we can use `types.MappingProxyType` instead

fandreuz: Also I think you don't need the typing annotation here, mypy will infer it

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    "tr_solver": "exact",
+    "loss": "linear",
+    "x_scale": "jac",
+    "gtol": 1e-8,
+    "xtol": 1e-8,
+    "ftol": 1e-8,
+}
+
+
+def __svht(
```
fandreuz, line 40: We don't use double underscore for very private things, one is enough

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    sigma_svd: np.ndarray,  # pylint: disable=unused-variable
```
fandreuz, line 41: Why is the `pyling` annotation needed?

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    rows: int,
+    cols: int,
+    sigma: float = None,
```
fandreuz, line 44: This should be annotated with `typing.Optional`

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+) -> int:
+    """
+    Determine optimal rank for svd matrix approximation,
+    based on https://arxiv.org/pdf/1305.5870.pdf.
+
+    :param sigma_svd:  Diagonal matrix of "enonomy" SVD as 1d array.
```
fandreuz, line 50:
```suggestion
    :param sigma_svd: Diagonal matrix of "enonomy" SVD as 1d array.
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    :return: Optimal rank.
+    :rtype: int
+    """
+
+    if len(sigma_svd.shape) != 1:
+        raise ValueError("Expected 1d array for diagonal matrix!")
+
+    beta = (
+        float(cols) / float(rows) if rows > cols else float(rows) / float(cols)
+    )
```
fandreuz, line 68:
```suggestion
    beta = max(rows, cols) / min(rows, cols)
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    if sigma is not None:
+        sigma = abs(sigma)
+        lambda_star = np.sqrt(
+            2 * (beta + 1)
+            + (8 * beta / (beta + 1 + np.sqrt(beta * beta + 14 * beta + 1)))
+        )
+        tau_star = (
+            lambda_star
+            * sigma
+            * np.sqrt(float(cols) if cols > rows else float(rows))
```
fandreuz, line 81:
```suggestion
            * np.sqrt(max(rows, cols))
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+        )
+    else:
+        median = np.median(sigma_svd)
+        beta_square = beta * beta
+        beta_cubic = beta * beta_square
+        omega = 0.56 * beta_cubic - 0.95 * beta_square + 1.82 * beta + 1.43
```
fandreuz, line 87: I propose to replace all this with `beta**2` and `beta**3`, this function is called only once

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+
+    return rank[-1] + 1
+
+
+def __compute_rank(
+    sigma_x: np.ndarray,
+    rows: int,
+    cols: int,
+    rank: Union[float, int],
+    sigma: float = None,
```
fandreuz, line 103: `typing.Optional`

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    :raises ValueError: ValueError is raised if rank is negative
+        (:math:`r < 0`).
+    :return: Optimal rank.
+    :rtype: int
+    """
+
+    if 0 < rank < 1:
+        sigma_x_square = np.square(sigma_x)
+        cumulative_energy = np.cumsum(sigma_x_square / sigma_x_square.sum())
+        __rank = np.searchsorted(cumulative_energy, rank) + 1
```
fandreuz, line 134: Local variable, no need for underscores

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    :type rank: Union[float, int], optional
+    :return: Diagonal matrix of eigenvalues
+        :math:`\boldsymbol{\Lambda}` as 1d array.
+    :rtype: np.ndarray
+    """
+
+    u_x, sigma_x, v_x_t = np.linalg.svd(
+        x_current, full_matrices=False, hermitian=False
+    )
+    __rank = __compute_rank(
```
fandreuz, line 170: Local variable, no need for underscores

fandreuz: It should receive a more suitable name to avoid conflict with the parameter `rank`

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    )
+
+    # columns of v need to be multiplicated with inverse sigma
+    sigma_inv_approx = np.reciprocal(sigma_x[:__rank])
+
+    a_approx = np.linalg.multi_dot(
+        [
+            u_x[:, :__rank].conj().T,
+            x_next,
+            sigma_inv_approx.reshape(1, -1) * v_x_t[:__rank, :].conj().T,
```
fandreuz, line 181:
```suggestion
            sigma_inv_approx[None] * v_x_t[:__rank].conj().T,
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+
+class __OptimizeHelper:  # pylint: disable=too-few-public-methods
+    """
+    Helper Class to store intermediate results during the optimization.
+    """
+
+    __slots__ = ["phi", "phi_inv", "u_svd", "s_inv", "v_svd", "b_matrix", "rho"]
+
+    def __init__(self, l_in: int, m_in: int, n_in: int) -> None:
+        self.phi: np.ndarray = np.empty((m_in, l_in), dtype=np.complex128)
```
fandreuz, line 196: Typing annotations not needed

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+
+
+class __OptimizeHelper:  # pylint: disable=too-few-public-methods
+    """
+    Helper Class to store intermediate results during the optimization.
+    """
+
+    __slots__ = ["phi", "phi_inv", "u_svd", "s_inv", "v_svd", "b_matrix", "rho"]
+
+    def __init__(self, l_in: int, m_in: int, n_in: int) -> None:
```
fandreuz, line 195: `-> None` is not needed: https://mypy.readthedocs.io/en/latest/class_basics.html#annotating-init-methods

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+            u_x[:, :__rank].conj().T,
+            x_next,
+            sigma_inv_approx.reshape(1, -1) * v_x_t[:__rank, :].conj().T,
+        ]
+    )
+
+    return np.linalg.eigvals(a_approx)
+
+
+class __OptimizeHelper:  # pylint: disable=too-few-public-methods
```
fandreuz, line 188: One underscore is enough

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    def __init__(self, l_in: int, m_in: int, n_in: int) -> None:
+        self.phi: np.ndarray = np.empty((m_in, l_in), dtype=np.complex128)
+        self.u_svd: np.ndarray = np.empty((m_in, l_in), dtype=np.complex128)
+        self.s_inv: np.ndarray = np.empty((l_in,), dtype=np.complex128)
+        self.v_svd: np.ndarray = np.empty((l_in, l_in), dtype=np.complex128)
+        self.b_matrix: np.ndarray = np.empty((l_in, n_in), dtype=np.complex128)
+        self.rho: np.ndarray = np.empty((m_in, n_in), dtype=np.complex128)
+
+
+def __compute_dmd_rho(
```
fandreuz, line 204: One underscore is enough

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    :rtype: np.ndarray
+    """
+
+    __alphas = np.zeros((alphas.shape[-1] // 2,), dtype=np.complex128)
+    __alphas.real = alphas[: alphas.shape[-1] // 2]
+    __alphas.imag = alphas[alphas.shape[-1] // 2 :]
+
+    phi = np.exp(np.outer(time, __alphas))
+    __u, __s, __v_t = np.linalg.svd(phi, hermitian=False, full_matrices=False)
+    __idx = np.where(__s.real != 0.0)[0]
```
fandreuz, line 238: Maybe this comparison should be done with `np.abs(...) > eps`? For some suitable `eps`

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    __s_inv = np.zeros_like(__s)
+    __s_inv[__idx] = np.reciprocal(__s[__idx])
+
+    rho = data - np.linalg.multi_dot([__u, __u.conj().T, data])
+    rho_flat = np.ravel(rho)
+    rho_out = np.zeros((2 * rho_flat.shape[-1],), dtype=np.float64)
+    rho_out[: rho_flat.shape[-1]] = rho_flat.real
+    rho_out[rho_flat.shape[-1] :] = rho_flat.imag
+
+    opthelper.phi = phi
```
fandreuz, line 248: Notice that `rho_out` can be computed based on `rho`, which is contained in `OptimizeHelper`. Therefore, I would make this method return an `OptimizeHelper`, which should be initialized in this function rather than requesting it as a parameter. `OptimizeHelper` should become a `frozen` dataclass: https://docs.python.org/3/library/dataclasses.html

pydmd/varprodmd.py
fandreuz: As discussed below, please make this a dataclass. And `Helper` is not appropriate, I would prefer something like `OptimizationData`

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+        d_phi_j = time * opthelper.phi[:, j]
+        __outer = np.outer(d_phi_j, opthelper.b_matrix[j, :])
+        __a_j = __outer - np.linalg.multi_dot(
+            [opthelper.u_svd, opthelper.u_svd.conj().T, __outer]
+        )
+        __g_j = np.linalg.multi_dot(
+            [
+                opthelper.u_svd * opthelper.s_inv.reshape((1, -1)),
+                np.outer(
+                    opthelper.v_svd[j, :].conj(), d_phi_j.conj() @ opthelper.rho
```
fandreuz, line 306:
```suggestion
                    opthelper.v_svd[j].conj(), d_phi_j.conj() @ opthelper.rho
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    """
+
+    __alphas = np.zeros((alphas.shape[-1] // 2,), dtype=np.complex128)
+    __alphas.real = alphas[: alphas.shape[-1] // 2]
+    __alphas.imag = alphas[alphas.shape[-1] // 2 :]
+    jac_out = np.zeros((2 * np.prod(data.shape), alphas.shape[-1]))
+
+    for j in range(__alphas.shape[-1]):
+        d_phi_j = time * opthelper.phi[:, j]
+        __outer = np.outer(d_phi_j, opthelper.b_matrix[j, :])
```
fandreuz, line 298:
```suggestion
        __outer = np.outer(d_phi_j, opthelper.b_matrix[j])
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+        __a_j = __outer - np.linalg.multi_dot(
+            [opthelper.u_svd, opthelper.u_svd.conj().T, __outer]
+        )
+        __g_j = np.linalg.multi_dot(
+            [
+                opthelper.u_svd * opthelper.s_inv.reshape((1, -1)),
```
fandreuz, line 304:
```suggestion
                opthelper.u_svd * opthelper.s_inv[None])),
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+                  np.ndarray,
+                  np.ndarray,
+                  np.ndarray]
+    """
+
+    u_r, s_r, v_r_t = np.linalg.svd(data, full_matrices=False)
+    __rank = __compute_rank(s_r, data.shape[0], data.shape[1], rank)
+    u_r = u_r[:, :__rank]
+    s_r = s_r[:__rank]
+    v_r = v_r_t[:__rank, :].conj().T
```
fandreuz, line 374:
```suggestion
    v_r = v_r_t[:__rank].conj().T
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+    data_out = v_r.conj().T * s_r.reshape((-1, 1)) if use_proj else data
```
fandreuz, line 375:
```suggestion
    data_out = v_r.conj().T * s_r[:, None] if use_proj else data
```

pydmd/varprodmd.py
```diff
@@ -0,0 +1,952 @@
...
+
+    # trapezoidal derivative approximation
+    y_out = (data_out[:, :-1] + data_out[:, 1:]) / 2.0
+    dt_in = time[1:] - time[:-1]
+    z_out = (data_out[:, 1:] - data_out[:, :-1]) / dt_in.reshape((1, -1))
```
fandreuz, line 380:
```suggestion
    z_out = (data_out[:, 1:] - data_out[:, :-1]) / dt_in[None]
```

fandreuz (commented):
Hi @greinerth, thanks for the PR! I started reviewing it, in general it's mostly about code style. In particular:
- I would like to see much less underscores (they increase noise and hide the problem of choosing better variable names);
- `.reshape(1, -1)` should become `[None]`
- `.reshape(-1, 1)` should become `[:, None]`
Please fix these in the whole MR so I can start the next review round :)

pydmd/varprodmd.py
greinerth: Done!

pydmd/varprodmd.py
greinerth: Okay, all the functions are now updated according to your suggestion ;)

pydmd/varprodmd.py
greinerth: Good catch, that is indeed unnecessary. It is now removed.

pydmd/varprodmd.py
greinerth: Done!

pydmd/varprodmd.py
greinerth: Done!

pydmd/varprodmd.py
greinerth: Done! Also, variable " __outer" is now substituted by "outer".

pydmd/varprodmd.py
greinerth: Done!

pydmd/varprodmd.py
greinerth: Done!

pydmd/varprodmd.py
greinerth: Done!

greinerth:
Hey @sichinaga, @mtezzele, @fandreuz, I wanted to ask how we could proceed with the module ;)

mtezzele:
Hi @greinerth, I am sorry for the delay but with Thanksgiving last week everything got delayed. 
We discussed also internally about this PR and we plan to merge it and see later on how to better integrate it with BOPDMD.
When @fandreuz is ok with the modifications we can proceed with the merge. 
Thanks for your patience.

mtezzele:
@fandreuz can we merge this or should we wait?

tests/test_varprodmd.py
```diff
@@ -0,0 +1,259 @@
+"""
```
fandreuz, line 1: Hi @greinerth, could you please apply here the same modifications you've done on the main code? Thanks

fandreuz:
Hi @greinerth, I see there are still some open comments. Could you please clarify if you've already addressed them in the code. Thanks!

tests/test_varprodmd.py
greinerth: Code style is updated now ;)

pydmd/varprodmd.py
```diff
@@ -0,0 +1,838 @@
...
+        # overwrite for lazy sorting
+        if isinstance(self._sorted_eigs, bool):
+            if self._sorted_eigs:
+                self._sorted_eigs = "auto"
+
+        if isinstance(self._sorted_eigs, str):
+            if self._sorted_eigs == "auto":
+                eigs_real = self._eigenvalues.real
+                eigs_imag = self._eigenvalues.imag
+                __eigs_abs = np.abs(self._eigenvalues)
```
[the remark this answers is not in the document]
greinerth: The double underscore is now a single underscore.

pydmd/varprodmd.py
greinerth: This is unfortunately not feasible as scipy's nonlinear least squares optimizer requires this function to compute a [vector of residuals](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html, which corresponds to parameter `fun` of the optimizer. This function is also used to avoid redundant computations since most entities are re-used in the Jacobian computation. The only way to achieve that is by manipulating the attributes of the `OptimizeHelper` class.

pydmd/varprodmd.py
greinerth: updated now

pydmd/varprodmd.py
greinerth: The annotations are gone now.

pydmd/varprodmd.py
greinerth: It is now removed.

fandreuz:
Looks good for me @mtezzele @ndem0, thanks @greinerth !

PR #473, opened 2023-11-08
`plot_summary()` improvements

sichinaga:
Hello everyone!

I've incorporated some new features into the `plot_summary()` function. I'll summarize them here in order from most crucial to least crucial. Let me know what you think! As always, I'm happy to make any necessary changes. :)

- `plot_summary()` **now uses time information stored in the given PyDMD model in order to more accurately compute discrete-time and continuous-time eigenvalues**, denoted as $\lambda$ and $\omega$ respectively. Before this update, the function always computed these values under the assumption that $\Delta t = 1$. But strictly-speaking, the relationship between these eigenvalues is $\omega = \log(\lambda) / \Delta t$, so in the event that one fits their data with $\Delta t \neq 1$, the function would compute and plot incorrect continuous-time eigenvalues. This has now been fixed.
    - On a similar note, since BOP-DMD always computes continuous-time eigenvalues and is able to accept snapshots that are unevenly sampled in time, there can be times where there is no uniform time-step $\Delta t$ and hence no discrete-time eigenvalues to plot. `plot_summary()` now detects this and doesn't plot the discrete-time eigenvalues if this is true.
    - The x-axis labels and y-axis labels for the eigenvalue plots are also now more descriptive.

- `plot_summary()` **now accepts the boolean parameter** `continuous`**, which allows users to specify if the eigenvalues stored in the given PyDMD model are continuous-time or discrete-time.** Before, the function always assumed that the eigenvalues were discrete-time, since they often are. However this isn't true in the event that a user fits their DMD model using the data matrix $\mathbf{X}$ and its derivative data $\mathbf{Y} = \dot{\mathbf{X}}$.

- `plot_summary()` now automatically detects if oscillations or changes in the DMD dynamics are extremely small (i.e. if the dynamics are approximately constant relative to the average value). **If the dynamics are approximately constant, the function will automatically re-adjust the window of the y-axis to more accurately reflect the constant nature of the dynamics, as opposed to emphasizing small and ultimately insignificant variations in the dynamics.** Time information is now incorporated into the x-axis of the dynamics plots as well.

- There are now **more parameters for plot customization.** Please see the method header and documentation for the full list of new parameters and their descriptions.

### DIFF
--- a/pydmd/plotter.py
+++ b/pydmd/plotter.py
@@ -538,8 +538,14 @@ def plot_summary(
     filename=None,
     order="C",
     figsize=(12, 8),
-    mode_colors=("r","b","g","gray"),
-    **kwargs,
+    dpi=200,
+    main_colors=("r", "b", "g", "gray"),
+    max_plot=50,
+    max_marker_size=10,
+    vmax_scale=0.9,
+    mode_cmap="bwr",
+    mode_color="tab:orange",
+    dynamics_color="tab:blue",
 ):
     """
     Generate a 3x3 summarizing plot that contains the following components:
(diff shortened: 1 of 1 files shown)


pydmd/plotter.py
```diff
@@ -532,30 +534,50 @@ def plot_snapshots_2D(
 
 def plot_summary(
     dmd,
+    continuous=False,
```
fandreuz, line 537: Since the number of arguments here is exploding, I propose to make everything after `dmd` a keyword-only argument. So the function signature would look like `def plot_summary(dmd, *, ...)`. This is a breaking change, but I think it's acceptable

fandreuz:
Thanks @sichinaga, looks goot to me!

pydmd/plotter.py
sichinaga: Added! I'm assuming this is what you meant @fandreuz

fandreuz: Yes exactly! But I would also like to get feedback from @mtezzele and @ndem0 before we break the interface :)

sichinaga:
Nice! I think this should be ready to merge whenever everyone is ready. :)

PR #485, opened 2024-01-18
Major HAVOK updates

sichinaga:
This PR contains a *massive* overhaul of the `HAVOK` module. Basically I've been working with HAVOK a fair amount in my own research and I thought it would be cool and helpful if I officially put my new and improved HAVOK code up on PyDMD.

Here are the new features introduced in this PR, roughly in order of their appearance in `havok.py`.

- The `HAVOK` module is now its own class as opposed to being a child of another PyDMD class.
This has been done for extra flexibility for the module.

- The `HAVOK` module now features several new input parameters:

    - `delays`: this is just the new name for `d`.

    - `lag`: the number of time-steps between each row of the Hankel matrix.
    This will give users more freedom in how they define their Hankel matrices.

    - `num_chaos`: the number of chaotic forcing terms to use in the HAVOK fit. This will allow users to model $n$-dimensional data sets that embed multiple chaotic forcing signals. In other words, `HAVOK` now supports $n$-dimensional data rather than just time-series data.

    - `structured`: whether to perform sHAVOK [(reference)](https://doi.org/10.1098/rspa.2021.0097) or traditional HAVOK.
    In other words, the `HAVOK` module now supports the sHAVOK algorithm.

    - `lstsq`: the method to use for the HAVOK fit if a DMD method isn't provided.
    Either a least-squares fit or the pseudo-inverse will be used.

    - `dmd`: a user-defined PyDMD instance to be used for the HAVOK fit. This parameter along with the previous parameter allow users to more clearly and explicitly define their desired method for performing the HAVOK fit.

- `hankel` and `dehankel` tools are also now available to users. Both functions are tested and will use the `delays` and `lag` parameters to build and undo Hankel matrices. (We might want to put these in `utils.py` eventually, but for now, the `lag` is just a fun little extra tool that users can mess around with if they want to).

- All SVD information can now be accessed from the `HAVOK` module. This is helpful in the event that users wish to view the time-delay modes or the singular value spectrum of the Hankel matrix.

- The `HAVOK` module now features a `compute_threshold` function, which can be used to help users plot and compute threshold values for the forcing terms that they compute with HAVOK.

- The `HAVOK` module now has its own `plot_summary` function (since the general `plot_summary` function does not produce meaningful results for HAVOK models).

This PR also contains **new tests** for these new features, and a **new tutorial** that walks through a basic HAVOK analysis that uses several of these new tools. The tutorial doesn't highlight *all* of these new features for the sake of conciseness, though I am thinking of highlighting the remaining more complicated features in another tutorial.

I know this PR is *very* long, so please review this PR at your leisure everyone. It has taken me a while to put this all together though and I hope these contributions are helpful to the PyDMD community. I will of course be on standby to answer any questions or address any feedback that may pop up. Thank you all in advance. :)

Best,
Sara Ichinaga


sichinaga:
Will fix the buggy tests in the next commit... 😅

pydmd/havok.py
```diff
@@ -1,276 +1,831 @@
...
+        if m < m_min:
+            raise ValueError(
+                "Not enough snapshots provided for "
+                f"{self._delays} delays and lag {self._lag}. "
+                f"Please provide at least {m_min} snapshots."
+            )
+
+        Hm = m - ((self._delays - 1) * self._lag)
+        H = np.empty((n * self._delays, Hm))
+        for i in range(self._delays):
```
fandreuz, line 306: Correct me if I'm wrong, but we have already something to compute the Hankel matrix, right? We should use that instead of re-implementing it, unless I'm missing something

pydmd/havok.py
```diff
@@ -1,276 +1,831 @@
...
+
+        # Fit a Gaussian to the forcing values.
+        mu, std = norm.fit(forcing)
+        gauss = norm.pdf(hx, mu, std)
+        gauss /= np.sum(gauss)
+
+        if isinstance(p, int):
+            # Use Gaussian intersection.
+            a = gauss - hy
+            ind_signchange = np.where(np.sign(a[:-1]) - np.sign(a[1:]) != 0)[0]
```
fandreuz, line 567:
```suggestion
            sgn = np.sign(a)
            ind_signchange = np.where(sgn(a[:-1]) * sgn(a[1:]) > 0)[0]
```

fandreuz:
Thanks @sichinaga! I don't have a deep understanding of the theory, I've left some stylistic comments.

pydmd/havok.py
sichinaga: Hi @fandreuz !

So the reason why I made a new implementation is because this implementation uses both a `delays` parameter and a `lag` parameter (the other implementation only uses a `delays` parameter). Basically, `lag` controls the time-skip that occurs between the rows of the Hankel matrix -- it's just a little extra feature and it might be helpful to let users play around with it when applying HAVOK.

We might want to put these in `utils.py` eventually, but for now, the lag is just a fun little extra tool that users can mess around with if they want to. :)

fandreuz: Ok I see, indeed I would like to merge the two versions so we keep only the most general one. It can be another though. Thanks for the explanation @sichinaga !

sichinaga:
Update: All of the tests seem to be passing now and the only issue seems to be some Codacy errors.

Personally, I think most if not all of these Codacy issues are pretty minor and can slide. In regards to the "medium" Codacy issues, the "unused argument t" issue cannot really be resolved because the function needs that header for `solve_ivp`. I also don't really think it's worth fixing the "Redefining name x from outer scope" issue for readability reasons.

sichinaga:
Oh yes @PyDMD/pydmd-devs, sorry for not really clearly stating this earlier, but I think this is okay to merge, unless we'd like to review it further before advancing. :)

PR #506, opened 2024-03-03
BOP-DMD UQ Tools

sichinaga:
Hi everyone!

It was great to see you all at SIAM UQ24! :)

My apologies for putting up a PR so soon after the conference, but this PR contains the UQ metric plotting tools that I demoed in the SIAM UQ24 minitutorial session -- I just thought it would be nice to turn these into actual plotting tools that are attached to the `BOPDMD` module so that users can seamlessly plot their UQ metrics.

This PR includes the following plotters for the `BOPDMD` class:
- A mode UQ metric plotter:
  - Plots the mode averages and standard deviations side-by-side.
  - Supports 1-D and 2-D data, as well as time-delay results.
  - Subplot layout is also customizable.
- An eigenvalue UQ metric plotter:
  - Plots the eigenvalue averages in the complex plane on top of circles that denote 1 and 2 standard deviations.
  - Users can plot true eigenvalues inside the plot for reference.
  - Axes configurations and limits are also customizable.
  
Here are some example figures and their respective calls:

```
bopdmd.plot_mode_uq(modes_shape=(ny, nx), cols=3, figsize=(14, 8), dpi=100)
```
image

 ```
bopdmd.plot_mode_uq(x=x, d=2, cols=4, figsize=(12, 5))
```
image

```
bopdmd.plot_eig_uq(
    eigs_true=[2.3j, -2.3j, 2.8j, -2.8j],
    flip_axes=True,
    figsize=(6, 2),
    dpi=200,
    draw_axes=True,
)
```
image

### CODE REVIEW

**Comment by klapo** [comment]
Maybe not the place to chime in on this, but with the plotter tools, would it make sense to add an option to omit one of the conjugate pairs when those are expected? In this example modes 3 and 5 contain no useful information when it comes to presenting results. An optional flag would allow you to check that the complex conjugate pairs behave as expected (e.g., `plot_complex_pair=True`) and then omit them (e.g. `plot_complex_pair=False` when generating figures for publication).

**Comment by sichinaga** [comment]
> Maybe not the place to chime in on this, but with the plotter tools, would it make sense to add an option to omit one of the conjugate pairs when those are expected? In this example modes 3 and 5 contain no useful information when it comes to presenting results. An optional flag would allow you to check that the complex conjugate pairs behave as expected (e.g., `plot_complex_pair=True`) and then omit them (e.g. `plot_complex_pair=False` when generating figures for publication).

@klapo This is a good idea. This in addition to allowing users to either pick specific modes and/or pick the number of modes to plot. I'll implement this soon.

File: pydmd/bopdmd.py [hunk 1]
```diff
@@ -1390,3 +1391,193 @@ def forecast(self, t):
...
+        # Collapse the results across time-delays.
+        if d > 1:
+            nd, r = modes.shape
+            modes = np.average(modes.reshape(d, nd // d, r), axis=0)
+            modes_std = np.average(modes_std.reshape(d, nd // d, r), axis=0)
+
+        # Define the subplot grid.
+        rows = 2 * int(np.ceil(modes.shape[-1] / cols))
+        plt.figure(figsize=figsize, dpi=dpi)
+        all_inds = np.arange(rows * cols).reshape(rows, cols)
```
fandreuz, line 1464: I'm confused about what you're trying to achieve here, could you please expand `avg` and `std` to make their role more clear?

pydmd/bopdmd.py
```diff
@@ -1390,3 +1391,193 @@ def forecast(self, t):
...
+        avg_inds = all_inds[::2].flatten()
+        std_inds = all_inds[1::2].flatten()
+
+        for i, (mode, mode_std) in enumerate(zip(modes.T, modes_std.T)):
```
fandreuz, line 1468: `enumerate` of `zip` of `.T` is a pretty long pipeline :)
I propose to use a simple `for i in range(...)` and extract what you need manually

pydmd/bopdmd.py
sichinaga: Are you referring to the conversion of `all_inds` to `avg_inds` and `std_inds`?

The reason for this part of the code is to figure out where all of the subplots should go. 

The intended behavior is that the resulting plot contains the average modes with their standard deviations plotted right below the averages. Users can also specify the number of columns of subplots. This way, a user can plot a 2 x 4 plot with `cols=4` like in the 2nd sample figure, or they could instead plot a 4 x 4 plot with `cols=2`, where the averages would occupy the 1st and 3rd rows of subplots and the standard deviations would occupy the 2nd and 4th rows.

Hence `avg_inds` contains the subplot numbers for where the average modes are gonna go, i.e. the 1st row of subplots, the 3rd row, etc. and `std_inds` does the same, but for the standard deviations.

Apologies for the confusion. I'll add more comments and documentation.

fandreuz: Thanks @sichinaga

sichinaga:
Hey everyone!

I'm finally back to wrap this PR up. Thanks again to everyone who looked over it.

So @fandreuz I added a few clarifying comments to hopefully help explain the confusing subplot formatting, and I made the for loop header a bit less confusing as well.

@klapo I also added the parameters `plot_modes` and `plot_conjugate_pairs` arguments so that in addition to the previous sample figures, you can now produce plots like the following (note that I'm using a different DMD rank for these results for greater contrast between the resulting modes, which is why these modes are different from the previous figure):

```
bopdmd.plot_mode_uq(modes_shape=(ny, nx), cols=3, figsize=(14, 8), plot_modes=6)
```
image

```
bopdmd.plot_mode_uq(modes_shape=(ny, nx), cols=3, figsize=(14, 5), plot_modes=3)
```
image

```
bopdmd.plot_mode_uq(modes_shape=(ny, nx), cols=3, figsize=(14, 5), plot_modes=3, plot_conjugate_pairs=False)
```
image

```
bopdmd.plot_mode_uq(modes_shape=(ny, nx), cols=3, figsize=(14, 5), plot_modes=(0, 3, 6))
```
image

sichinaga:
On that note, I think I'm ready to merge as long as everyone else is satisfied with the code! 😄

fandreuz:
Good for me, thanks @sichinaga !

mtezzele:
I proceed with the merge. @sichinaga please in the future do not merge the pydmd master branch into your local branch, do instead a proper rebase (fetch + rebase + force push), thanks.

PR #511, opened 2024-03-19
fix outdated docstrings

Doublonmousse:
There was some outdated doctest in the `utils.py` dating from the last refactor (where a function name was changed but not the associated doctest). I made sure that 

```py
python -m doctest utils.py
```
was successful.

Are doctest not part of the current testing CI ? Seems like that could be fixed with a change of configuration for `pytest` (see here : https://docs.pytest.org/en/7.1.x/how-to/doctest.html)

### DIFF
--- a/pydmd/utils.py
+++ b/pydmd/utils.py
@@ -277,19 +277,20 @@ def pseudo_hankel_matrix(X: np.ndarray, d: int) -> np.ndarray:
     :Example:
 
         >>> a = np.array([[1, 2, 3, 4, 5]])
-        >>> _hankel_pre_processing(a, d=2)
+        >>> pseudo_hankel_matrix(a, d=2)
         array([[1, 2, 3, 4],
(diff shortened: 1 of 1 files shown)


mtezzele:
Thank you so much @Doublonmousse! @fandreuz I think we can add this to the CI, what do you think?

mtezzele:
@Doublonmousse please run `black` on the modified file.

Doublonmousse:
Okay. I've reformatted the file using `black`.

Seems like using `doctestplus` you can run the following command 
```py
pytest pydmd/ --doctest-only
```
to only test doctest.

Though it doesn't seem like it's the only doctest failing doing that ...

image

Doublonmousse:
Argh ... seems all `.. code-block:: python` are tested with this option, and without surprises all such code that suppose that you have `dmd` as a dmd object before running them fail

PR #515, opened 2024-03-30
Lagged dmdc

nmank:
The changes I've added.

1. Added time lag as a variable to the DMDc class so the snapshot matrix can be split into X and Y using lags other than just 1

2. Modified the create_system_without_B function in test_dmdc to generate data with unknown B and time lag in the snapshots.

Do you have a systematic way to do testing for this addition? I'm happy help out!


pydmd/dmdc.py
```diff
@@ -193,7 +193,7 @@ class DMDc(DMDBase):
     :type svd_rank_omega: int or float
     """
 
-    def __init__(self, svd_rank=0, tlsq_rank=0, opt=False, svd_rank_omega=-1):
+    def __init__(self, svd_rank=0, tlsq_rank=0, opt=False, svd_rank_omega=-1, lag = 1):
```
mtezzele, line 196: please no spaces around the equal sign

tests/test_dmdc.py
```diff
@@ -12,15 +12,16 @@ def create_system_with_B():
     return {"snapshots": snapshots, "u": u, "B": B}
 
 
-def create_system_without_B():
+def create_system_without_B(lag = 1):
```
mtezzele, line 15: no spaces around the equal sing.

mtezzele:
@nmank could you please add some documentation about the new parameter please?

nmank:
Thanks! I will, I realized that a more principled way to do this is to pass in X and Y to the fit function and remove a 'lag' parameter. I've also been writing new tests for this function. I need to get the reconstruction test to work. Once my modification and the tests are working, should I just make another pull request or modify this one?

mtezzele:
You can keep this PR, so the discussion remains on a single page and it is easier to follow. But do it as you prefer, closing this PR and opening a new one is also fine.

pydmd/dmdc.py
```diff
@@ -268,7 +272,10 @@ def reconstructed_data(self, control_input=None):
             [self.modes, np.diag(eigs), np.linalg.pinv(self.modes)]
         )
 
-        data = [self.snapshots[:, 0]]
+        data = []
+        for i in range(self._lag):
+            data.append(self.snapshots[:, i])
```
mtezzele, line 277: Please use list comprehension

pydmd/dmdc.py
```diff
@@ -298,13 +305,18 @@ def fit(self, X, I, B=None):
...
-        self._controlin = np.atleast_2d(np.asarray(I))
+        n_samples = self.snapshots.shape[-1]
 
-        n_samples = self.snapshots.shape[1]
-        X = self.snapshots[:, :-1]
-        Y = self.snapshots[:, 1:]
+        if self._lag < 1:
+            raise ValueError(
+            f"Time lag must be positive."
+                )
```
mtezzele, line 316: Move this in the constructor.

tests/test_dmdc.py
```diff
@@ -26,6 +26,21 @@ def create_system_without_B():
     return {"snapshots": snapshots, "u": u, "B": B, "A": A}
 
 
+def create_system_without_B_lag(lag = 1):
```
mtezzele, line 29: space around the =

mtezzele (commented):
Since you added a ValueError please add a test that is able to catch that.

mtezzele:
@nmank just a curiosity: can you describe a use case where you need this lag?

nmank:
> @nmank just a curiosity: can you describe a use case where you need this lag?

I was using it to extract modes of variability from global surface air temperature under different emissions scenarios. So like building DMDc using X from times 0 to N-10 and Y from times 10 to N. See the last few slides [here](https://maths4dl.ac.uk/wp-content/uploads/2024/04/MankovichSlides.pdf).

The problem with this is that we did not get eigenvalues on the unit circle. :( I've spent the last while trying to get tutorial 1 working on the data with OptDMD or BopDMD. But I can't seem to the alternating projection method to work.

PR #519, opened 2024-04-05
Fix plotter.plot_summary() method which doesn't plot modes correctly when given snapshots_shape for 2D snapshots

karlic-luka:
**Note: I copy-pasted my PR description:**
---------------------------------------------------------

Official documentation says that `snapshots_shape `parameter is supposed to be: **(width, height).**

Taking that into account, plotting of the 2D modes inside `plot_summary()` method was wrong.
I will explain it using my use-case. I am using DMD for background extraction. 

1. These are my original snapshots (320x240 images, where width=320) plotted using `plotter.plot_snapshots_2D` method
original_snapshots

2. Modes using `plotter.plot_summary()`are complete "junk"- they aren't explainable at all (however, dimensions are right - 320x240, where width is 320)
modes_original_method

3. Modes using fixed `plotter.plot_summary()` from my local PyDMD library
image

We can see that now modes are explainable - i.e. first mode is actually a background, other modes are some other features.

The problem occurred during reshaping of snapshots into original shape because it was done _column-wise_ instead of _row-wise_. Order was default (C-style), but we were actually doing F-style ordering.

P.S. methods `plotter.plot_modes_2D` and `plotter.plot_snapshots_2D` are doing the similar thing, but they are correctly taking into account that provided parameter `snapshots_shape` is **(width, height)**, not **(height, width)** and they are not using `snapshots_shape` parameter to reshape.

P.P.S. I used the exact same approach from those 2 methods to fix this one

### DIFF
--- a/pydmd/plotter.py
+++ b/pydmd/plotter.py
@@ -899,7 +899,7 @@ def plot_summary(
     if len(snapshots_shape) == 2:
         if y is None:
             y = np.arange(snapshots_shape[1])
-        ygrid, xgrid = np.meshgrid(y, x)
+        xgrid, ygrid = np.meshgrid(x, y)
 
     # PLOTS 4-6: Plot the DMD modes.
     for i, (ax, idx) in enumerate(zip(mode_axes, index_modes)):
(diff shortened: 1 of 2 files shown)


mtezzele:
@sichinaga, since you spent a lot of time on this method, could you please have a look at this PR please?

sichinaga:
I’ll take a look asap.

sichinaga:
Hi @karlic-luka! Thank you for the pull request. 😄 

I think the code looks great! Thank you for catching this and for looking into this so throughly.

@mtezzele I'm good to merge!

PR #522, opened 2024-04-18
Readme updates

sichinaga:
Hi everyone!

I've added some readme updates to catch our documentation up with more current figures and developments.

I will summarize the major changes below in order in which they appear:

- Converted the "Using PyDMD" section into a "Quickstart Guide" section. The section now contains (1) a very basic DMD example, (2) how to do basic preprocessing, and (3) how to build a more complex BOP-DMD model.
- Added sample `plot_summary` output to go with the DMD application example already in the readme.
- Added a "Citing PyDMD" section, which contains the JOSS paper and the arXiv paper.
- Added a few major DMD references that we were missing previously.
- Updated the "Current PyDMD Capabilities" figure to reflect the completion of COSTS and randomized preprocessing.
- Added the back-end API structure figure from the paper to the developer tutorial to make it more accessible for devs.


mtezzele:
Thank you, @sichinaga! I was wondering if it is time to get rid of the gif and the noisy data figure. Maybe we can mention the tutorials in the new quickstart section and move the plotting figure there as well. In this way we can remove entirely the examples and tutorials section. What do you think?

sichinaga:
> Thank you, @sichinaga! I was wondering if it is time to get rid of the gif and the noisy data figure. Maybe we can mention the tutorials in the new quickstart section and move the plotting figure there as well. In this way we can remove entirely the examples and tutorials section. What do you think?

@mtezzele Yeah this would work! I think either format is fine, but you're right that the two sections are a bit redundant, especially since I do mention the tutorials and the documentation in the quickstart section. I think I'll go ahead with this change unless any of the @PyDMD/pydmd-devs have any qualms.

sichinaga:
@PyDMD/pydmd-devs As an update, the quickstart and tutorial/examples section have been consolidated into one.

I've also added a `plot_summary` code block as a part of the quickstart tutorial.

There are a few more tutorials and small module updates that I'd like to make for easier user onboarding, but that is work for another PR. I think these changes should be ready for merge unless there are more issues and/or comments!

README.md
```diff
@@ -91,44 +91,78 @@ and then install the package in [development mode](https://setuptools.pypa.io/en
...
+# X = (n, m) numpy array of time-varying snapshot data.
+dmd.fit(X)
+
+# Plot a summary of the key spatiotemporal modes.
+plot_summary(dmd)
+```
+
+PyDMD modules can also be wrapped with data preprocessors if desired. These wrappers will preprocess the data and postprocess data reconstructions automatically.
+```python3
+from pydmd import DMD
```
fandreuz, line 114: Small note, we could consider this snippet as a continuation of the previous one, so we would not need to import `DMD`. It's better to focus on what we want to highlight, no strong opinion though

fandreuz:
Thanks for the improvements @sichinaga, looks good to me!

README.md
sichinaga: Hey @fandreuz! I ultimately opted for the current code snippets because I felt that being able to see where everything was imported from within the same snippet added a little bit of extra clarity. I would like to keep the current format as long as it's alright with you!

PR #524, opened 2024-04-26
Some necessary `BOPDMD` maintenance and updates

sichinaga:
After using `BOPDMD` a bit for my own work, I've been able to catch a few hiccups in the code. This PR seeks to remedy a variety of bugs and meet several user needs related to the module. Here is a summary of the major changes (in order in which they appear):
- The parameter `mode-prox` has been added, whose functionality is similar to `eig_constraints`, only the parameter acts on the modes as opposed to the eigenvalues. It's a quick and simple addition that will allow users to apply thresholding or smoothing to their BOP-DMD modes. This parameter is still in progress, and its addition is merely setup for its eventual incorporation as a major parameter.
- Non-converged bagging trials are now kept by default to prevent (1) users needing to manually tinker with the `tol` parameter to achieve convergence, and (2) users inadvertently triggering eternal runtimes. The functionality of the `bag_maxfail` parameter has been updated to achieve this.
- The `"conjugate_pairs"` eigenvalue constraint code has been de-bugged. The function previously assumed a specific ordering of the eigenvalues, which was true for many datasets and cases, but I found it wasn't true for all. Now the code doesn't make any assumptions on the ordering of the eigenvalues and should hence be more robust and consistent.
- A somewhat minor update, but the code within the `variable_projection` function has been compartmentalized more and made more explicit. Some computations were previously lumped together into vague function calls, which made the components of the variable projection difficult to decipher. There was also a conflation of terms for quantities such as the objective and the error. These issues are now remedied.
- Mode shaping routine for the `plot_mode_uq` function has been fixed in accordance with [this recent PR](#519).

**I have also added a tutorial that very quickly highlights several of these new, decently major functionalities, along with some other basic but very major features of the `BOPDMD` module.** The intent for this tutorial was that it would be somewhat along the lines of a quickstart tutorial that very briefly directs new users to major features and information. The tutorial however grew very quickly, so I felt that calling it a user manual would be more appropriate. But with that in mind, I felt that it would be appropriate to highlight it as a different type of tutorial, though I am open to feedback on how we should approach this!


pydmd/bopdmd.py
```diff
@@ -72,13 +72,22 @@ class BOPDMDOperator(DMDOperator):
...
+        full input data.
+    :type mode_prox: function
     :param bag_warning: Number of consecutive non-converged trials of BOP-DMD
         at which to produce a warning message for the user. Default is 100.
         Use arguments less than or equal to zero for no warning condition.
     :type bag_warning: int
     :param bag_maxfail: Number of consecutive non-converged trials of BOP-DMD
-        at which to terminate the fit. Default is -1, no stopping condition.
-    :type bag_maxfail: int
+        at which to terminate the fit. Set this parameter to infinity for no
```
fandreuz, line 87: Correct me if I'm wrong, we used to select `no stopping condition` when user gives `bag_maxfail=-1`, and now this behavior is selected when we `bag_maxfail=infty`. I have two comments:
- This is a breaking change, we should be careful. Can we achieve the same result in some different way, e.g. a new parameter?
- Basic python users may not know how to represent infinity, I propose to replace `infinity` in the function docs with e.g. `np.inf`

pydmd/bopdmd.py
```diff
@@ -285,23 +296,29 @@ def _push_eigenvalues(self, eigenvalues):
...
-                eig_pair = eigenvalues[i : i + 2]
-                real_comp = np.mean(eig_pair.real)
-                imag_comp = np.mean(np.abs(eig_pair.imag))
-                new_eigs[i] = real_comp + 1j * imag_comp
-                new_eigs[i + 1] = real_comp - 1j * imag_comp
+                ind_0 = np.argmin(np.abs(eigenvalues.imag))
+                new_eigs[ind_0] = eigenvalues[ind_0].real
+                unassigned_inds.remove(ind_0)
+            # Assign complex conjugate pairs until all eigenvalues are paired.
+            while len(unassigned_inds) != 0:
```
fandreuz, line 307:
```suggestion
            while unassigned_inds:
```

pydmd/bopdmd.py
```diff
@@ -285,23 +296,29 @@ def _push_eigenvalues(self, eigenvalues):
...
+                ind_1 = unassigned_inds.pop()
+                eig_1 = eigenvalues[ind_1]
+                # Get the index of the eigenvalue that's closest to being the
+                # complex conjugate of the randomly selected eigenvalue.
+                ind_2 = np.argmin(np.abs(eigenvalues - np.conj(eig_1)))
+                eig_2 = eigenvalues[ind_2]
+                unassigned_inds.remove(ind_2)
+                # Average their real and imaginary components together.
+                a = 0.5 * (eig_1.real + eig_2.real)
+                b = 0.5 * (abs(eig_1.imag) + abs(eig_2.imag))
```
fandreuz, line 318: I propose to always NumPy `abs` just to get consistent behavior (plus I would some very small performance gain)

pydmd/bopdmd.py
```diff
@@ -285,23 +296,29 @@ def _push_eigenvalues(self, eigenvalues):
         if "conjugate_pairs" in self._eig_constraints:
             num_eigs = len(eigenvalues)
             new_eigs = np.empty(num_eigs, dtype="complex")
+            unassigned_inds = set(np.arange(num_eigs))
```
fandreuz, line 299: Using `set` for an ordered list of unique values is somewhat unusual. Also, in the `pop()` below we are relying on an undocumented mechanism to make sure it has [constant complexity](https://stackoverflow.com/a/66592754), but in general `Set.pop()` is supposed to be `O(N)`. I propose to use this as a standard NumPy array, and using `-1` as a placeholder for taken indexes instead of removing them from the data structure.

pydmd/bopdmd.py
```diff
@@ -285,23 +296,29 @@ def _push_eigenvalues(self, eigenvalues):
...
+            # Assign complex conjugate pairs until all eigenvalues are paired.
+            while len(unassigned_inds) != 0:
+                # Randomly grab the next unassigned eigenvalue.
+                ind_1 = unassigned_inds.pop()
+                eig_1 = eigenvalues[ind_1]
+                # Get the index of the eigenvalue that's closest to being the
+                # complex conjugate of the randomly selected eigenvalue.
+                ind_2 = np.argmin(np.abs(eigenvalues - np.conj(eig_1)))
+                eig_2 = eigenvalues[ind_2]
+                unassigned_inds.remove(ind_2)
```
fandreuz, line 315: If you accept my comment above about removing the `set` conversion, you can replace this with `unassigned_inds[ind_2] = -1`

pydmd/bopdmd.py
```diff
@@ -285,23 +296,29 @@ def _push_eigenvalues(self, eigenvalues):
...
-            for i in eig_pair_inds:
-                eig_pair = eigenvalues[i : i + 2]
-                real_comp = np.mean(eig_pair.real)
-                imag_comp = np.mean(np.abs(eig_pair.imag))
-                new_eigs[i] = real_comp + 1j * imag_comp
-                new_eigs[i + 1] = real_comp - 1j * imag_comp
+                ind_0 = np.argmin(np.abs(eigenvalues.imag))
+                new_eigs[ind_0] = eigenvalues[ind_0].real
+                unassigned_inds.remove(ind_0)
+            # Assign complex conjugate pairs until all eigenvalues are paired.
```
fandreuz, line 306: @sichinaga one question (I'm ignorant about the method implemented here so I could definitely be missing something). When you remove an index from `unassigned_inds`, you prevent it from being selected as a future `ind_1` in L309. However, you always use the full content of `eigenvalues` (which is never updated) to find the appropriate `ind_2`, thus an eigenvalue could be potentially selected multiple times as `ind_2`. Is this correct?

pydmd/bopdmd.py
```diff
@@ -581,29 +608,29 @@ def compute_residual(alpha):
                 (rhs_top[:IA], np.zeros(IA, dtype="complex")), axis=None
             )
 
-            def step(_lambda, rhs, scales_pvt, ij_pvt):
+            def step(_lambda, scales_pvt=scales_pvt, rhs=rhs, ij_pvt=ij_pvt):
```
fandreuz, line 611: Do we need this to be defined here? Could it be a private (maybe static) method of the class?

fandreuz:
Thanks @sichinaga! Left a bunch of comments

klapo:
Hi @sichinaga, I just wanted to highlight some behavior of BOP-DMD I noticed which may be relevant for future directions of the `mode-prox` keyword. Sometimes the phase information for $\omega$ is represented by the sign of $\phi$ (i.e., the sign of $\phi$ modes are used by BOP-DMD to fit phase offsets). This really only became noticeable because I was looking at collections of BOP-DMD fits through time in the mrCOSTS routine. I'm excited to play with this new functionality!

pydmd/bopdmd.py
sichinaga: @fandreuz To answer your question, no. For clarification, the intended behavior is as follows:

The mission is to figure out which eigenvalues in `eigenvalues` are approximately complex conjugate pairs. Once we identify these approximate pairs, we enforce that they are _actually_ complex conjugate pairs. I do this by defining `unassigned_inds`, which is a set that is initialized to contain all of the indices that correspond with entries of `eigenvalues`. The order of these indices should not matter in theory. The goal is that we uniquely pair up indices in `unassigned_inds`. Once 2 indices in `unassigned_inds` are identified as a pair, they are removed from `unassigned_inds` so that they are never paired again.

- If we are dealing with an odd number of eigenvalues, we start by identifying the eigenvalue with the smallest imaginary part. We take this eigenvalue to be real and we remove its index from `unassigned_inds`.
- We then pop an index from `unassigned_inds` since it needs to be paired. This is `ind_1`, which corresponds with `eig_1`. 
- We then look through `eigenvalues` and find the index of the eigenvalue that is closest to being the complex conjugate of `eig_1`. This is now `ind_2`, which corresponds with `eig_2`. We then call `unassigned_inds.remove(ind_2)` so that both `ind_1` and `ind_2` are removed from `unassigned_inds`.

The confusion might be coming from that last bullet. In practice, as long as you're dealing with eigenvalues that are approximately conjugate pairs, `ind_2` ends up being something that hasn't been removed from `unassigned_inds` yet, and everything works out. If you're instead dealing with a completely mixed bag of numbers, what is identified as `ind_2` might no longer be in the `unassigned_inds` set, hence causing `unassigned_inds.remove(ind_2)` to trigger an error (which is kind of my intention in a way).

If we think this might cause issues, maybe I can throw a try catch block in instead?

pydmd/bopdmd.py
sichinaga: I use a set here because the order of the entries of `unassigned_inds` doesn't matter -- the data structure is used for the sole purpose of keeping track of index values that still need attention. (I explain more [here](https://github.com/PyDMD/PyDMD/pull/524#discussion_r1584021081).) Do you still think we should remove the set usage given my method explanation?

pydmd/bopdmd.py
sichinaga: The purpose of this method was always to avoid a small amount of redundancy. This method is only called in two lines. I don't think it needs to be a private method because of this.

pydmd/bopdmd.py
sichinaga: I've restored the original functionality of `bag_maxfail`.

Now, a boolean parameter called `remove_bad_bags` has been introduced instead.

pydmd/bopdmd.py
fandreuz: Ah it's ok, it's just that with this style we get one level of indentation for no visible code simplicity gain. Please resolve the thread if you like it this way

pydmd/bopdmd.py
fandreuz: Thanks!

pydmd/bopdmd.py
fandreuz: > I think it's also worth keeping in mind that eigenvalues will typically only contain svd_rank-many entries, so maybe the O(n) runtime isn't the worst? Tradeoffs to consider I guess?

Sure, but since we use `pop` in a loop it could become an `O(N^2)` if it wasn't for the efficient CPython implementation.

No strong opinion on this, I commented mainly because of the `set(np.arange(...))`

pydmd/bopdmd.py
fandreuz: > hence causing unassigned_inds.remove(ind_2) to trigger an error (which is kind of my intention in a way).

Ah I see, it's clear now. Thanks for the explanation!

I don't like throwing an exception in non-exceptional situations. I would rather check proactively if `ind_2 in unassigned_inds`, and then throw a `ValueError` (or even a custom exception type) containing a nice error message

sichinaga: Sounds good!

I've added the functionality, along with a few extra comments to clarify what exactly is going on inside the method.

Let me know if I should add anything else here! :)

mtezzele:
@sichinaga please next time rebase the branch wrt the master branch. Do not merge PyDMD:master into the current branch.
Also no need for all these commits, at least some of them can be marked as fixes. 
I proceed with accepting this PR.

PR #530, opened 2024-05-09
Additional tests for `BOPDMD` and `LANDO` modules

sichinaga:
Hi everyone!

In preparation for our journal submission, I've nearly doubled the number of tests for `BOPDDMD` and `LANDO` to improve the coverage of our package. It seems that these two modules were the most lacking in terms of coverage (each seem to have ~70% coverage prior to this update), which is why I paid special attention to these modules. I've also added some fit checking for some `BOPDMD` attributes, as to help users avoid strange behaviors when calling these attributes.

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -214,8 +214,6 @@ class BOPDMDOperator(DMDOperator):
                 "Set parameter compute_A = True to compute A."
             )
             raise ValueError(msg)
-        if self._A is None:
-            raise ValueError("You need to call fit before")
         return self._A
 
     @property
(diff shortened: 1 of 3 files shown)


mtezzele:
Thank you for this PR! Just a minor correction: when testing the plotting generation please add `plt.close()` at the end of the corresponding test. In the past after a certain number of open figures the tests fail.

sichinaga:
Thanks for the tip @mtezzele! I've added a call to `plt.close()` in both of the plotting tests. I also made a small function tweak to get rid of the Codacy issue. Let me know if anything else needs to be done to this PR! :)

PR #532, opened 2024-05-27
Fixed bug in building convolution windows for mrCOSTS

klapo:
Hi all,

I found a small bug in how mrCOSTS builds it convolution windows and performed window rounding. The long story short was that I didn't correctly adapt the original MATLAB indexing for python. This pull request fixes that issue.

Summary of changes:
- Convolution windows had MATLAB style indexing and not python. As a result the windows did not peak at the right place.
- Similarly, the window rounding was asymmetric.
- Causes slight reduction in the reconstructed error in the toy data example.
- `b` amplitudes now fall along PSD of the input data, fixing a lingering issue in the mrCOSTS fit and giving it more diagnostic power.
- Fixed the mrCOSTS tests not correctly tearing down.

Thanks for taking a look!
Karl

### DIFF
--- a/pydmd/costs.py
+++ b/pydmd/costs.py
@@ -325,13 +325,11 @@ class COSTS:
 
         lv_kern = (
             np.tanh(
-                corner_sharpness
-                * np.arange(1, window_length + 1)
-                / window_length
+                corner_sharpness * np.arange(0, window_length) / window_length
             )
             - np.tanh(
                 corner_sharpness
-                * (np.arange(1, window_length + 1) - window_length)
+                * (np.arange(0, window_length) - window_length - 1)
                 / window_length
             )
             - 1
@@ -354,7 +352,7 @@ class COSTS:
         """
         recon_filter_sd = window_length / 8
         recon_filter = np.exp(
-            -((np.arange(window_length) - (window_length + 1) / 2) ** 2)
+            -((np.arange(window_length) - (window_length - 1) / 2) ** 2)
             / recon_filter_sd**2
         )
         return recon_filter
--- a/tests/test_mrcosts.py
+++ b/tests/test_mrcosts.py
@@ -149,8 +149,8 @@ expected_n_components = 2
 
 # Define the expected error in the reconstructions.
 expected_global_error = 0.053
-expected_lf_error = 0.12
-expected_hf_error = 0.19
+expected_lf_error = 0.10
+expected_hf_error = 0.17
 expected_transient_error = 0.3
 
 # Fit mrCOSTS for testing
@@ -360,7 +360,7 @@ def test_plot_local_time_series():
         _ = mrc.plot_local_time_series(0, 0, data=data.T)
 
 
-def tear_down():
+def test_tear_down():
     """Remove the files generated in `test_netcdf`"""
     file_list = glob.glob("*tests*.nc")
     for f in file_list:
--- a/pydmd/mrcosts.py
+++ b/pydmd/mrcosts.py
@@ -1041,14 +1041,8 @@ class mrCOSTS:
 
             # Convolve each windowed reconstruction with a gaussian filter.
             # Std dev of gaussian filter
-            recon_filter_sd = mrd.window_length / 8
-            recon_filter = np.exp(
-                -(
-                    (np.arange(mrd.window_length) - (mrd.window_length + 1) / 2)
-                    ** 2
-                )
-                / recon_filter_sd**2
-            )
+            recon_filter = mrd.build_kern(mrd.window_length)
+
             omega_classes = omega_classes_list[n_mrd]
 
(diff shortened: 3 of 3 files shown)


tests/test_mrcosts.py
```diff
@@ -360,7 +360,7 @@ def test_plot_local_time_series():
         _ = mrc.plot_local_time_series(0, 0, data=data.T)
 
 
-def tear_down():
+def test_tear_down():
```
fandreuz, line 363: Renaming this function will cause pytest to discover the tear down routine as a test: https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery

pydmd/costs.py
```diff
@@ -325,13 +325,11 @@ def calculate_lv_kern(window_length, corner_sharpness=None):
 
         lv_kern = (
             np.tanh(
-                corner_sharpness
-                * np.arange(1, window_length + 1)
-                / window_length
+                corner_sharpness * np.arange(0, window_length) / window_length
```
fandreuz, line 328:
```suggestion
                corner_sharpness * np.linspace(0, 1, window_length, endpoint=False)
```

tests/test_mrcosts.py
klapo: Hi @fandreuz how could I trigger the tear down without calling it test?

fandreuz: The best solution is to use a temporary directory: https://docs.pytest.org/en/6.2.x/tmpdir.html
All files produced by the test should be put inside this directory, which is fed into each test function via a fixture. Let me know if you need help to set this up :)

klapo:
Just as a public note, the lv_kern function still returns an asymmetric kerning (`(np.arange(0, window_length) - window_length - 1)` should be `(np.arange(0, window_length) - window_length + 1)` ). Unfortunately, making the window symmetric seems to reduce the fit quality in testing, so I am just accepting that this is a weird idiosyncrasy of the method for the time being.

pydmd/costs.py
klapo: Accepted

tests/test_mrcosts.py
klapo: I believe I implemented this correctly now using the `tmp_path` fixture.

klapo:
I managed to break everything.

The tutorial tests failing is Notebook 2 (Advanced DMD) because scipy.integrate.trapz wasn't found. The most recent release of scipy removed `.trapz` in favor or `.trapezoid` (https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html). Do we want to pin the scipy version to be >=1.14 and update the call to the non-deprecated version?

klapo:
I think the unit tests broke because I did `os.chdir(tmp_path)` in the mrCOSTS unit tests. Then the other tests were pointing to the wrong directory when it came time for them to load the test data. I don't have the internet connection to download the test extras so I'm testing this directly here on github instead of locally.

mtezzele:
> I managed to break everything.
> 
> The tutorial tests failing is Notebook 2 (Advanced DMD) because scipy.integrate.trapz wasn't found. The most recent release of scipy removed `.trapz` in favor or `.trapezoid` (https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html). Do we want to pin the scipy version to be >=1.14 and update the call to the non-deprecated version?

Thank you for pointing this out. @fandreuz @ndem0 @sichinaga Should we add a try/else since version 1.14 is the latest one? I don't want to limit the current users. IMHO we can impose the version after 1 or 2 new versions. What do you think?

fandreuz:
@mtezzele I propose to keep only the new version, `scipy.integrate.trapezoid` has been around for 4 years: https://github.com/scipy/scipy/blob/v1.6.0/scipy/integrate/_quadrature.py

fandreuz:
> > @mtezzele I propose to keep only the new version, `scipy.integrate.trapezoid` has been around for 4 years: https://github.com/scipy/scipy/blob/v1.6.0/scipy/integrate/_quadrature.py
> 
> Ok, so we can enforce scipy >= version_of_4_years_ago instead of the current one, which is quite new. What do you think?

Yes looks good! You want to do it? Otherwise I'll open a PR

mtezzele:
> > > @mtezzele I propose to keep only the new version, `scipy.integrate.trapezoid` has been around for 4 years: https://github.com/scipy/scipy/blob/v1.6.0/scipy/integrate/_quadrature.py
> > 
> > 
> > Ok, so we can enforce scipy >= version_of_4_years_ago instead of the current one, which is quite new. What do you think?
> 
> Yes looks good! You want to do it? Otherwise I'll open a PR

Go ahead please! Thank you very much!

mtezzele:
Let me know if I can merge this PR

klapo:
I believe so. Thank you to everyone!

PR #536, opened 2024-07-19
Migrate `setup.py` to `pyproject.toml`, and CalVer

fandreuz:
- Full migration to `pyproject.toml`: https://peps.python.org/pep-0621/
- I propose to adopt CalVer to simplify our versioning: https://calver.org/
- Bumped version for several GitHub actions


fandreuz:
@PyDMD/pydmd-devs any comment?

mtezzele:
@fandreuz I like it, it seems way much easier to manage and it aligns better with our monthly versions. If @ndem0 is ok we can merge this PR. Thanks!

PR #554, opened 2024-10-28
Refactor pre/post processing implementation

### LINKED ISSUE #553: Pre-processing for DMDc
**Describe the bug**
Hello! Are the pre-processing methods not supported for DMDc models? Even a simple zero-mean pre-processing doesn't work; returning the error:

**Output**
> File ~\anaconda3\envs\dmdc\Lib\site-packages\pydmd\preprocessing\pre_post_processing.py:103 in _pre_processing_fit
>     self._pre_processing(self._state_holder, *args, **kwargs)
> 
> TypeError: _pre() got multiple values for argument 'axis'

Looks like this is because the pre-processing function only expects one dataset (the state snapshots), but is receiving two datasets (the state snapshots and input snapshots) from the DMDc model. Is this a bug or an unsupported feature?

**To Reproduce**
```
import numpy as np
from pydmd import DMD,DMDc
from pydmd.preprocessing import zero_mean_preprocessing

stateData = np.random.rand(10,100)
inputData = np.random.rand(1,100)

# works fine
dmd = DMD(svd_rank=0)
dmd = zero_mean_preprocessing(dmd)
dmd.fit(stateData)

# does not work
dmdc = DMDc(svd_rank=0)
dmdc = zero_mean_preprocessing(dmdc)
dmdc.fit(stateData,inputData[:,:-1])
```

Same problem/error with the other pre-processing methods.

Environment packages I have installed: [environment.zip](https://github.com/user-attachments/files/17546805/environment.zip) [same issue on both v1.0.1 and v1.0.0]

fandreuz:
Fixes #553

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -1369,7 +1369,7 @@ class BOPDMD(DMDBase):
         # rank-truncated SVD of Y.
         Y = (ux1 + ux2) / 2
         # Element-wise division by time differences. w/o large T
-        Z = (ux2 - ux1) / (self._time[1:] - self._time[:-1])  
+        Z = (ux2 - ux1) / (self._time[1:] - self._time[:-1])
         U, s, V = compute_svd(Y, self._svd_rank)
         S = np.diag(s)
 
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -1712,16 +1712,16 @@ class BOPDMD(DMDBase):
 
         if flip_axes:
             eigs = self.eigs.imag + 1j * self.eigs.real
-            plt.xlabel("$Im(\omega)$")
-            plt.ylabel("$Re(\omega)$")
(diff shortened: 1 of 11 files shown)


fandreuz:
Hi @klapo, `tests/test_mrcosts.py::test_reconstructions` is failing (see the failing jobs), I think you've recently worked there. Do you know what's the problem?

mtezzele:
Thank you, @fandreuz ! As soon as the TypeErrors are fixed I'm ok with this PR.

klapo:
@fandreuz Have these tests been failing before or did this just start? It is difficult for me to see what could have caused the failing test in this commit. I do see a commit from October 6th which touches something pretty important to mrCOSTS though.

In any case, I have an update for mrCOSTS that will require changing these tests in the works. I'm happy to loosen up the testing requirements, especially since whatever happened here improved the fit.

PR #549, opened 2024-10-11
Bug fix for the variable projection algorithm of BOPDMD 

dsj976:
There is a small bug in the variable projection algorithm implemented in the `BOPDMD` class. This bug was causing the algorithm to fail to converge for certain "complex" signals (e.g. composed of three sine waves).

This is the [relevant line](https://github.com/duqbo/optdmd/blob/eff57045ab4085ae2b6a1a2decb583deeb1be488/src/varpro2.m#L272) in the original Matlab code.

I have made a [repo](https://github.com/dsj976/test-opt-dmd) were you can test these changes on a pre-generated dataset. Read the `README.md` for instructions.

PR #563, opened 2024-12-17
Eigenvalue constraints for bopdmd, improvements for mrCOSTS, consistency with variable projection bug fix

klapo:
Dear PyDMD devs,

This pull request spans several topics so it is a bit complicated. The main focus of this branch was originally improving the eigenvalue constraints in bopdmd. The `conjugate_pairs` eigenvalue constraint now explicitly forces conjugate pairs instead of raising an error when the solver creates an ambiguous set of conjugate pairs. I used failures found in my testing with real world data to inform the tests implemented for the `conjugate_pairs` eigenvalue constraints in case you need to see examples of why this change is necessary.

Simultaneous with this branch are several overall improvements to mrCOSTS functionality, mostly related to better handling of keywords between modules. The biggest change to this component from a user's perspective is improved fits at the edge of the time domain and to the highest frequency components through improvements to the reconstruction convolution.

The master branch accepted a change to the variable projection algorithm in the bopdmd module. Unfortunately, there was substantial interaction between that bug fix and the main topic of this pull request. I documented the issues with this bug fix in the discussion on Pull Request #549 . The best solution I've found so far is to just simply accept that one can only impose very loose limits on the real eigenvalues. If too strict of limits on the real component of eigenvalues are placed, the solver _can_, but not necessarily will, fail in some hard to understand ways. 

In response to the solver failures I originally tried to improve the variable projection algorithm. But, these limitations generally did not work, just creating a harder to understand algorithm. I've opted to simply discard the algorithm changes and proceed with this state of the algorithm with limitations on what real eigenvalue constraints can do. To better inform users of this limitation I am planning an update for the mrCOSTS and BOPDMD tutorials.

I tried to make the history more reflective of the current state through aggressive fix up of commit messages and removing the alterations to the variable projection algorithm. It is likely I overlooked something though.

Thank you for your help! I am looking forward to getting this fix out.

Karl

### DIFF
--- a/pydmd/bopdmd.py
+++ b/pydmd/bopdmd.py
@@ -712,8 +712,10 @@ class BOPDMDOperator(DMDOperator):
             q_out, djac_out, j_pvt = qr(
                 djac_matrix, mode="economic", pivoting=True
             )
-            ij_pvt = np.zeros(IA, dtype=int)
-            ij_pvt[j_pvt] = np.arange(IA, dtype=int)
+            ij_pvt = np.arange(IA)
+            ij_pvt = ij_pvt[j_pvt]
+            # ij_pvt = np.zeros(IA, dtype=int)
+            # ij_pvt[j_pvt] = np.arange(IA, dtype=int)
             rjac[:IA] = np.triu(djac_out[:IA])
             rhs_top = q_out.conj().T.dot(rhs_temp)
             scales_pvt = scales[j_pvt[:IA]]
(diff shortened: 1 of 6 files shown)


klapo:
Hi all, I somehow really gorped up the git history as a bunch of commits are now doubled. My intent was to squash the commits down considerably, but instead i just made the log much worse.

I also somehow re-introducing a long-since fixed bug.  I apologize for this mess of a pull request.

I will close this commit and simply open a pull request on the branch without the messed up git history and old bug. Hopefully that one will be easier to evaluate!

Issue #558, opened 2024-11-22
numpy 2.0 compatibility


ilyapopov:
Currently, PyDMD is explicitly specified as incompatible with numpy 2.0 https://github.com/PyDMD/PyDMD/blob/1f6db1f15a6d07b435c6e6d25387c21f8c891074/pyproject.toml#L7

It would be nice if that limitation was lifted. 

Ref: https://numpy.org/doc/stable/numpy_2_0_migration_guide.html

fathurwithyou:
Yes, I tried installing using pip and it showed error "ModuleNotFoundError: No module named 'numpy.char'" because of the numpy 1.26.4. To solve this problem, I upgrade the version to be 2.2.1. I think the lib should be updated.

PR #567, opened 2025-01-14
Conj pairs constraints + var-pro + mrCOSTS improvements

klapo:
Dear PyDMD devs,

This pull request spans several topics so it is a bit complicated. The main focus of this branch was originally improving the eigenvalue constraints in bopdmd. The `conjugate_pairs` eigenvalue constraint now explicitly forces conjugate pairs instead of raising an error when the solver creates an ambiguous set of conjugate pairs. I used failures found in my testing with real world data to inform the tests in case you need to see examples of why this change is necessary.

Simultaneous with this branch are several overall improvements to mrCOSTS functionality. The biggest change to this component from a user's perspective is improved fits at the edge of the time domain and to the highest frequency components through improvements to the reconstruction convolution.

The master branch accepted a change to the variable projection algorithm in the bopdmd module. Unfortunately, there was substantial interaction between that bug fix and the main topic of this pull request. I documented the issues with this bug fix in the discussion on Pull Request https://github.com/PyDMD/PyDMD/pull/549 . The best solution I've found so far is to just simply accept that one can only impose very loose limits on the real eigenvalues. If too strict of limits on the real component of eigenvalues are placed (e.g. `stable` eigenvalue constraints), the solver can, but not necessarily will, fail in some hard to understand ways.

In response to the solver failures I originally tried to improve the variable projection algorithm. But, these limitations generally did not work, just creating a harder to understand algorithm. I've opted to simply discard the algorithm changes and proceed with this state of the algorithm with limitations on what real eigenvalue constraints can do. To better inform users of this limitation I am planning an update for the mrCOSTS and BOPDMD tutorials.

Thank you for your help! I am looking forward to getting this fix out.

Karl


pydmd/bopdmd.py
```diff
@@ -283,6 +292,148 @@ def _varpro_opts_warn(self):
...
+            # current state of the solver eigenvalues.
+            msg = (
+                "Trouble pairing conjugate pairs. \n"
+                f"Pair indices = {pair_indices}"
+                f"eigenvalues = {eigenvalues}"
+            )
+            raise ValueError(msg)
+
+        eigenvalues = np.copy(new_eigs)
+        return eigenvalues
```
mtezzele, line 435: `return np.copy(new_eigs)`

pydmd/bopdmd.py
```diff
@@ -283,6 +292,148 @@ def _varpro_opts_warn(self):
...
+            ind_1, ind_2 = np.nonzero(diff_array == np.nanmin(diff_array))
+            ind_1 = ind_1[0]
+            ind_2 = ind_2[0]
+
+            eig_1 = eigenvalues[ind_1]
+            eig_2 = eigenvalues[ind_2]
+
+            unassigned_inds.remove(ind_2)
+            unassigned_inds.remove(ind_1)
+            diff_array[ind_1, :] = np.nan
```
mtezzele, line 399: I think you can get rid of the `:` when they are in the last position. There are several cases in this file.

pydmd/costs.py
```diff
@@ -317,29 +356,50 @@ def calculate_lv_kern(window_length, corner_sharpness=None):
...
 
-        # Higher = sharper corners
         if corner_sharpness is None:
             corner_sharpness = 16
 
-        lv_kern = (
-            np.tanh(
-                corner_sharpness
-                * np.linspace(0, 1, window_length, endpoint=False)
+        if kern_method == "kern":
```
mtezzele, line 369: Could you please add some comments on the formulas used in this if block?

klapo: These were actually part of a test I was doing and then discarded. So I simply removed several of the options.

mtezzele:
@fandreuz @sichinaga do you have any comments on this PR? Otherwise I will proceed with the merge right after merging PR 568.

mtezzele:
@klapo could you please rebase this PR? I cannot do it online because of some minor conflicts with the last merged PR.

klapo:
@mtezzele I just clicked merge to see what would happen and github said it worked...

pydmd/bopdmd.py
```diff
@@ -283,6 +292,147 @@ def _varpro_opts_warn(self):
...
+            if sign == 0.0:
+                diff = np.abs(np.abs(eigenvalues.imag) - omega.imag)
+                diff[ind] = np.nan
+            else:
+                same_sign_index = np.sign(eigenvalues.imag) == sign
+                opp_sign_eigs = np.copy(eigenvalues.imag)
+                opp_sign_eigs[same_sign_index] = np.nan
+                diff = np.abs(np.abs(opp_sign_eigs) - np.abs(omega.imag))
+
+        return diff
```
fandreuz, line 336: I propose an early return here:
```suggestion
            diff[ind] = np.nan
            return diff
 
        sign = np.sign(omega.imag)
        
        # Catch the edge case of the eigenvalues being exactly 0.
        if sign == 0.0:
            diff = np.abs(np.abs(eigenvalues.imag) - omega.imag)
            return diff[ind] = np.nan
            
        same_sign_index = np.sign(eigenvalues.imag) == sign
        opp_sign_eigs = np.copy(eigenvalues.imag)
        opp_sign_eigs[same_sign_index] = np.nan
        return np.abs(np.abs(opp_sign_eigs) - np.abs(omega.imag))
```

pydmd/bopdmd.py
```diff
@@ -283,6 +292,147 @@ def _varpro_opts_warn(self):
...
+            num_eigs_adj = num_eigs
+
+        # Determine if we have eigenvalues with opposite signs or if we are
+        # in a weird state where we might not be able to match eigenvalues.
+        if np.count_nonzero(eigenvalues.imag > 0) == np.count_nonzero(
+            eigenvalues.imag < 0
+        ):
+            absolute_diff = False
+        else:
+            absolute_diff = True
```
fandreuz, line 377:
```suggestion
        absolute_diff = np.count_nonzero(eigenvalues.imag > 0) != np.count_nonzero(
            eigenvalues.imag < 0
        )
```

pydmd/bopdmd.py
```diff
@@ -283,6 +292,147 @@ def _varpro_opts_warn(self):
...
+
+        for nomega, omega in enumerate(eigenvalues):
+            # Comparing just the imaginary components allows the
+            # conjugate pair identification to be insensitive to the real
+            # part.
+            diff_array[nomega] = self._diff_func(
```
fandreuz, line 383: `_diff_func` is a `staticmethod`, thus we should not use `self` to access it

klapo: @fandreuz I am a little confused because this is how other static methods are accessed throughout PyDMD (e.g., see `_test_kernel_inputs` in `edmd.py` and `compute_irank_svd` in `bopydmd.py`.

How should I access it otherwise?

klapo:
There are a number of failing tests. As best as I can discern numpy v2 crashes out mrCOSTS. The failing tests for numpy v1.26.4 are from other modules (LANDO).

The errors seems be at least in part because the solver chooses real eigenvalues that result in solutions going to infinity. My best guess is there is a weird interaction between the matrix inversions in the solver using numpy v2 and the strict conjugate pair requirement that causes the eigenvalue solutions to blow up. If this is the case, it would be similar to the errors I ran into when we did the variable projection bug fix a couple months ago. Unfortunately, I won't have time to look at this in the next week, but I will look into it after that.

Unless someone else sees an obvious cause/solution?

pydmd/bopdmd.py
fandreuz: The correct form is `ClassName.static_method_name` :
```suggestion
            diff_array[nomega] = BOPDMDOperator._diff_func(
```

It works because static methods are also accessible by instances of the class

fandreuz:
> There are a number of failing tests. As best as I can discern numpy v2 crashes out mrCOSTS. The failing tests for numpy v1.26.4 are from other modules (LANDO).
> 
> The errors seems be at least in part because the solver chooses real eigenvalues that result in solutions going to infinity. My best guess is there is a weird interaction between the matrix inversions in the solver using numpy v2 and the strict conjugate pair requirement that causes the eigenvalue solutions to blow up. If this is the case, it would be similar to the errors I ran into when we did the variable projection bug fix a couple months ago. Unfortunately, I won't have time to look at this in the next week, but I will look into it after that.
> 
> Unless someone else sees an obvious cause/solution?

Hi @klapo, thanks for looking into the failures. I don't have a full understanding of the problem, but I'd say it's something we should look into before merging this in since we support `numpy>=2` (#558 was merged a couple of weeks ago). 

I propose not to rush anything before we have the full picture, these failures on NumPy v2 look quite serious to me

klapo:
I found the cause of the errors. It turns out I was abusing a poorly defined behavior for `numpy.sign` in regards to complex numbers, which was corrected in version 2 (https://github.com/numpy/numpy/pull/25441). With v2 of numpy and my poor use of `numpy.sign` I was in essence just flipping the real and imaginary components, leading to all of the failing tests. 

The conjugate pair constraint now explicitly checks the sign of the imaginary component instead of the entire complex eigenvalue, which results in the original desired behavior.

pydmd/costs.py
```diff
@@ -70,6 +70,15 @@ class COSTS:
     :param reset_alpha_init: Flag specifying if the initial eigenvalue guess
         should be reset between windows.
     :type reset_alpha_init: bool
+    :param kern_method: Specifies if the fit should be made to the data
```
fandreuz, line 73: Could you please replace all `kern` with `kernel`? This does not buy us much in terms of readability

pydmd/costs.py
```diff
@@ -317,29 +356,49 @@ def calculate_lv_kern(window_length, corner_sharpness=None):
...
+                )
+                - 1
             )
-            - 1
-        )
+        elif kern_method == "flat":
+            # Do not apply a window kerning prior to fitting each window.
+            lv_kern = np.ones(window_length)
+        else:
+            raise ValueError("Invalid `kern_method` provided.")
```
fandreuz, line 396: Please print the value as part of the exc message

pydmd/costs.py
```diff
@@ -348,14 +407,33 @@ def build_kern(window_length):
...
         )
+        # Do not apply the kerning at the end of the time domain. This
+        # assists in fitting the last window of the decomposition and stops
+        # the propagation of errors at the edge of the time domain.
+        if direction == "forward":
+            recon_filter[(window_length // 2) :] = 1
+        # Do not apply the kerning at the beginning of the time domain. This
+        # assists in fitting the last window of the decomposition and stops
+        # the propagation of errors at the edge of the time domain.
+        elif direction == "backward":
```
fandreuz, line 434: Should we throw an exception if `direction` is anything else than those two possibilities?

fandreuz:
> I found the cause of the errors. It turns out I was abusing a poorly defined behavior for `numpy.sign` in regards to complex numbers, which was corrected in version 2 ([numpy/numpy#25441](https://github.com/numpy/numpy/pull/25441)). With v2 of numpy and my poor use of `numpy.sign` I was in essence just flipping the real and imaginary components, leading to all of the failing tests.
> 
> The conjugate pair constraint now explicitly checks the sign of the imaginary component instead of the entire complex eigenvalue, which results in the original desired behavior.

Thanks @klapo for tracking down the problem! I left some minor comments, otherwise this PR is good to go for me

pydmd/costs.py
klapo: I'm happy to include this, but I was not intending for this keyword to be exposed to the user. Would that intent modify how we approach this?

klapo:
Hi all, just checking if there is something I'm overlooking that is left to do for this pull request. It looks like it is done to me?

sichinaga:
Sorry I haven't been doing a great job about keeping up with this PR, but I looked over it now and this looks good to me! This is also probably my sign to fix up the LANDO tutorial in a future PR.

PR #579, opened 2025-05-14
Pre-post processing refactor

klapo:
Hi PyDMD devs,

Per our conversation, I was quite confused when looking at touching the pre_post_processing tools. I made some attempts to rename some variables and add comments to all methods in the hopes that these changes would be easier to understand both for myself and future devs.

In particular, I was keeping an eye open to exposing a larger surface area of the code to the processing hooks.

Thank you for your time!
Karl


klapo:
The failing tutorial tests seem to be related to this:
https://github.com/wso2/product-is/issues/23500

pydmd/preprocessing/pre_post_processing.py
```diff
@@ -43,83 +74,103 @@ class PrePostProcessingDMD(Generic[S]):
     def __init__(
         self,
         dmd: DMDBase,
-        pre_post_processing: PrePostProcessing[S] = PrePostProcessing(),
+        processing_hooks: PrePostProcessing[S] = PrePostProcessing(),
```
fandreuz, line 77: I think we're introducing too many names here, do we need to have `hooks`? We already have an abstraction of a pre/post processing action, which is `PrePostProcessing`. Having an instance called similarly to the class name `pre_post_processing` is a common practice, and the class has now a nice documentation telling what each function does.

pydmd/preprocessing/pre_post_processing.py
```diff
@@ -43,83 +74,103 @@ class PrePostProcessingDMD(Generic[S]):
...
     ):
         if dmd is None:
             raise ValueError("DMD instance cannot be None")
-        self._pre_post_processing = pre_post_processing
+        self._processing_hooks = processing_hooks
 
-        self._dmd = dmd
-        self._pre_processing_output: S | None = None
+        self._wrapped_dmd = dmd
+        self._processing_state: S | None = None
```
fandreuz, line 84: `state` seems to imply that the state can be kept after you're done with a given training dataset, `pre_processing_output` conveys the role of the field more effectively in my opinion

pydmd/preprocessing/pre_post_processing.py
```diff
@@ -43,83 +74,103 @@ class PrePostProcessingDMD(Generic[S]):
...
+        self._wrapped_dmd.modes_activation_bitmask = value
 
-    def _pre_processing_fit(self, *args, **kwargs):
-        X = PrePostProcessingDMD._extract_training_data(*args, **kwargs)
-        self._pre_processing_output, pre_processed_training_data = (
-            self._pre_post_processing.pre_processing(X)
+    def _fit_with_preprocessing(self, *args, **kwargs):
+        """
+        Runs pre-processing before calling the wrapped DMD's fit method.
+        Stores the pre-processing state for use during post-processing.
```
fandreuz, line 134: I think "state" is unclear here, isn't "output" a better fit to explain the algorithm?

pydmd/preprocessing/svd_projection.py
```diff
@@ -55,7 +55,5 @@ def pre_processing(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
         return projection_matrix, projection_matrix.T.dot(X)
 
     @override
-    def post_processing(
-        self, pre_processing_output: np.ndarray, Y: np.ndarray
-    ) -> np.ndarray:
-        return pre_processing_output.dot(Y)
+    def post_processing(self, state: np.ndarray, Y: np.ndarray) -> np.ndarray:
```
fandreuz, line 58: Let's keep this consistent with the definition of `PrePostProcessing`, if parameter names change here they should also be changed upstream

fandreuz:
Hi @klapo, thanks for the PR! It looks good to me. Is there anything specific we should look at, except for the naming changes?

pydmd/preprocessing/pre_post_processing.py
klapo: Again the issue is readability. If every variable is appended with "pre", "post", or "pre_post" one cannot assert at a glance where it is coming from because they all blend together.

Also, we need to keep the state after training, no? e.g., the mean when centering has to be kept around in order to do the reconstruction.

pydmd/preprocessing/svd_projection.py
klapo: Then I would argue to change them upstream because of the readability issue.

pydmd/preprocessing/pre_post_processing.py
klapo: Is the idea here that you are defining a state as something internal to the system and the output as something external to the system? I saw this as being something that is not externally visible since this transformation is never visible to the user.

mtezzele:
@klapo is there anything else to do for this PR? If not, please rebase and merge the PR.

